PR #2, opened 2021-11-08
Simplify pypi publish workflow trigger.

SpacemanPaul:
PYPI action wasn't triggering.  This part of the 1.0.5 release.

### DIFF
--- a/.github/workflows/publish-pypi.yml
+++ b/.github/workflows/publish-pypi.yml
@@ -1,11 +1,7 @@
 name: Publish to PyPI
 
 on:
-  workflow_run:
-    workflows: ["Run Code Checks"]
-    branches: [pypi/publish]
-    types:
-      - completed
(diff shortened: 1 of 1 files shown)


supermarkion (approved):
All good to me

Kirill888:
@supermarkion @SpacemanPaul this won't work! This workflow takes wheels from the other workflow and only runs if testing and packaging was successful. What was probably wrong in there before is branch condition, you want to push from `develop` not from `py-publish`

supermarkion:
> @supermarkion @SpacemanPaul this won't work! This workflow takes wheels from the other workflow and only runs if testing and packaging was successful. What was probably wrong in there before is branch condition, you want to push from `develop` not from `py-publish`

Hi, @Kirill888 . Thanks a lot for your information. We will check it.

PR #3, opened 2021-11-08
Tweak pypi publishing workflow

Kirill888:
Publish from develop branch after successful completion of code checks
This will only publish to pypi if version has been changed.
So no longer need to push to `pypi/publish` in order to make release, this "work around" only made sense in `odc-tools`.


SpacemanPaul (commented):
Looks great. Was thinking of doing something similar when I got around to it.

SpacemanPaul (approved):

PR #9, opened 2021-11-30
datetime iso8601 compatible

emmaai:
convert iso8601 compatible datetime string to `datetime.datetime`


SpacemanPaul (commented):
Ok, you will also need to 

1) add ciso8601 to `install_requires` in `setup.cfg`.
2) add ciso8601 to `docker/requirements.txt`

and if you want to be able to include it in Statistician docker builds, you will need to force a release by incrementing the version in `_version.py`.

alexgleith (approved):

PR #12, opened 2021-12-20
Revert "Add in UE filter for FC summaries"

alexgleith:
Reverts opendatacube/odc-stats#11


alexgleith:
Just checking we _DO_ want to make the proposed change.

PR #15, opened 2021-12-20
FC Percentiles Changes - Fix counts, make sum > 120 NODATA and clip to 0-100

alexgleith:
* Counts valid observations properly
* Sets pixels that sum to > 120 to NODATA
* Clips input data to between 0-100


alexgleith:
Hey @fangfy and @cbur24 this is ready for review.

Pixel counts look like this:

image


And over exposed areas went from this:

image

to this:

image

cbur24 (approved):

emmaai (changes requested):
I'm blocking this pr, because
1. filter by unmixing error 30, why it has to be 30? any evidence back up the value? how many false positive or neglected negative when thresholding by this value?
2. clipping to 100, I got the idea of having the output of fc percentile clipping to 100, don't get the idea of clipping the input from FC. 100 or 101 doesn't make too much difference in computing percentile tbh. And if something is 127, usually that pixel is no valid.
3. filter by sum to 120, again why 120? it's set by experimental results? or visually good enough results?
4. with all the changes, hard to understand what's the implication for the downstream product, especially for mangroves.

emmaai:
@alexgleith  please undo the merge. thanks.

alexgleith:
Hey @emmaai I don't want to undo the merge, but I'm happy to make further changes to make settings optional or to change thresholds.

emmaai:
> Hey @emmaai I don't want to undo the merge, but I'm happy to make further changes to make settings optional or to change thresholds.

Yes, if you can, please make those thresholds optional. Thanks.

PR #17, opened 2022-01-06
Mangroves plugin

emmaai:
Add mangroves plugin


odc/stats/plugins/mangroves.py
```diff
@@ -0,0 +1,113 @@
+"""
+Fractional Cover Percentiles
+"""
```
SpacemanPaul, line 3: File docstring needs to be updated.

odc/stats/plugins/mangroves.py
```diff
@@ -0,0 +1,113 @@
...
+        gdal.RasterizeLayer(target_ds, [1], source_layer, burn_values=[1])
+        return dask.array.from_array(band.ReadAsArray().reshape(array_shape), name=False)
+
+    def native_transform(self, xx):
+        """
+            no native transform required since
+            no reprojection in mangroves
+            return loaded data
+        """
+        return xx
```
SpacemanPaul: This is OK, but unnecessary as it is identical to the default implementation is provided by the base class.

odc/stats/plugins/mangroves.py
```diff
@@ -0,0 +1,113 @@
...
+
+    def fuser(self, xx):
+        """
+            no fuse required for mangroves either
+            return loaded data
+        """
+        return xx
```
SpacemanPaul, line 74: This is different to the default implementation.  Are you sure this is what you want?

SpacemanPaul (changes requested):
At the very least please change the file docstring to say "Mangroves" instead of "Fractional Cover".

Also please double check whether the fuser function is correct or if you actually want the _nodata_fuser() default implementation.   It might be unnecessary because you are doing masking in the reduce stage instead of the dataload stage.

Nice to see some tests though!

odc/stats/plugins/mangroves.py
emmaai: Yes, I'm sure. There is nothing to group, hence nothing to fuse.

odc/stats/plugins/mangroves.py
emmaai: deleted

SpacemanPaul (approved):

PR #18, opened 2022-02-07
Features needed

emmaai:
Changes in this PR

- read remote config file.
- added cli function of getting a list of grids covered by polygons in a given shape file.
- added cli option of printing version, easy for debugging and docker compiling, required the higher click version accordingly.
- moved `datasets-filter` and `apply-eodatasets3` options to config file, which can be overwritten by the command-line. Since these two options are shared the same among at least 3 products, it will reduce the trouble/mistakes in carrying them in each product.
- made `count_valid` band optional, we (DEA) don't have a clear definition or user case for the band.
- imported  `dc_load` from correct package.
- pumped up the version.


SpacemanPaul (approved):
LGTM

alexgleith (changes requested):
This all looks ok.

Can we please add tests for the CLI config loading?

alexgleith:
I think a test of the CLI tool that does spatial intersection would be a good idea too.

emmaai:
> This all looks ok.
> 
> Can we please add tests for the CLI config loading?

Done

emmaai:
> 

As for this, turned out mocking an esri shapefile is non-trivial. So I'm leaving it out for this iteration.

odc/stats/_text.py
```diff
@@ -81,6 +81,22 @@ def parse_yaml_file_or_inline(s: str) -> Dict[str, Any]:
     return result
 
 
+def load_yaml_remote(yaml_url: str) -> Dict[str, Any]:
+    """
+       Open a yaml file remotely and return the parsed yaml document 
+    """
+    import urllib
+    import fsspec
```
omad, line 89: `fsspec` should be added to `install_requires` in `setup.cfg`. Hopefully adding this exception doesn't cause any troubles, but there's a chance it will due to `aiobotocore`.

odc/stats/_text.py
```diff
@@ -81,6 +81,22 @@ def parse_yaml_file_or_inline(s: str) -> Dict[str, Any]:
     return result
 
 
+def load_yaml_remote(yaml_url: str) -> Dict[str, Any]:
+    """
+       Open a yaml file remotely and return the parsed yaml document 
+    """
+    import urllib
```
omad, line 88: This function has many problems. 
1. It would need to import `urllib.error` to access the `urllib.error.URLError` exception class.
2. However, testing now, `fsspec` raises a `FileNotFoundError` if it can't access a file, and `yaml` raises `yaml.scanner.ScannerError` if it can't parse the input, so there's no point trying to catch `URLError`.
3. There is no `{value}` in scope for the `f-string` on line 95.

I think it would be reasonable to skip the Exception handling entirely here, since there's no good way to resolve the error.

OR

Alternatively, we could catch the exceptions and log a message saying `f'unable to load {yaml_url}` and then have a bare `raise` to re-raise the current exception.

OR

Define a new  `class YamlLoadError(Exception)`, and `raise YamlLoadError(f'Unable to load {yaml_url}')`  from inside the `except`.

tests/test_text.py
```diff
@@ -29,6 +29,14 @@ def test_parse_yaml():
     assert set(o) == {"a", "b"}
 
 
+@pytest.mark.parametrize("yaml_url, expected", [("https://raw.githubusercontent.com/GeoscienceAustralia/dea-config/"\
```
omad, line 32: I'm not keen on tests that depend on external resources and URLs, but, I know it's kind of painful to do otherwise. So, okay...

alexgleith (approved):

omad (changes requested):
Looks mostly okay thanks Emma.

In future, could you please break this many changes into multiple PRs to make it easier to review and merge.

tests/test_text.py
alexgleith: We have an established pattern for handling this, actually. I wasn't going to say anything, but I agree here.

alexgleith: Here's the example of faking a URL with a local file: https://github.com/digitalearthafrica/deafrica-scripts/blob/main/deafrica/tests/test_chirps.py#L105

PR #20, opened 2022-02-09
Fix yaml remote loading

emmaai:
address @omad and @alexgleith  comments on yaml remote loading and unit testing in pr #18 
- catch all the exceptions and raise in general, add the test
- mock url with `pytest.httpserver` rather than loading remotely in unit test
- add `fsspec` in the requirments
- pump up the version again


tests/test_text.py
```diff
@@ -29,12 +29,20 @@ def test_parse_yaml():
...
+    """
+    )
+    content = load_yaml_remote(httpserver.url_for("/ga_ls_fc_pc_cyear_3.yaml"))
+    assert content["a"] == 3 and content["b"] == "foo"
+    try: 
+        content = load_yaml_remote(httpserver.url_for("/something_non_exist.yaml"))
+    except Exception as e:
+        pass
+    else:
+        assert False
```
alexgleith: This is not the right way to check that an exception is raised.

See: https://docs.pytest.org/en/6.2.x/assert.html#assertions-about-expected-exceptions

emmaai: thanks! fixed.

tests/test_text.py
```diff
@@ -29,12 +29,16 @@ def test_parse_yaml():
...
+def test_load_yaml_remote(httpserver):
+    httpserver.expect_request("/ga_ls_fc_pc_cyear_3.yaml").respond_with_data("""
+    a: 3
+    b: foo
+    """
+    )
+    content = load_yaml_remote(httpserver.url_for("/ga_ls_fc_pc_cyear_3.yaml"))
+    assert content["a"] == 3 and content["b"] == "foo"
+    with pytest.raises(FileNotFoundError, match=r".*something_non_exist.yaml"):
+        load_yaml_remote(httpserver.url_for("/something_non_exist.yaml"))
```
alexgleith, line 41: I think you should change this to use a fake URL, rather than `httpserver.url_for`, as you're catching the error of the library you're using to do the test, rather than the underlying function.

So this test is not actually helpful!

emmaai: the scenario is that the server is alive, but the file doesn't exist on the server since the server doesn't respond to the request of `/something_non_exist.yaml`. Is it fundamentally different than that the server does not exist with an arbitrary url? I'll expect either of them will raise `FileNotFoundError`.

alexgleith: If you do:

```
with fsspec.open("http://example.com/fake.yaml", mode="r") as f:
    y = next(yaml.safe_load_all(f))
```

it raises `ClientResponseError`.

emmaai: so that's `fsspec` version related behaviour, not url related.
If `fsspec >= 2021.11.*`, the error is
```
FileNotFoundError: http://example.com/fake.yaml
```
if `fsspec` is much older, the error is
```
ClientResponseError
```
for both cases.

That's my latest commit trying to address, in bumping `fsspec` version in both test and install

alexgleith: Ok, I see. Thanks!

omad (approved):
Looks good, thanks Emma.

PR #21, opened 2022-02-18
fix thumbnail generating with dask

emmaai:
The issue: the thumbnail generating function triggered the dask computation unintentionally so that the monitoring mechanism doesn't work as it was designed to be.
The solution: keep the dask graph unpacked until `client.compute` by `dask.delayed`

I've tried my best to stay on the best practice of `dask.delayed` and not to tamper with the dask graph. Hence the shuffle and change of functions and code.

P.S.: it's a short term solution to enable us into production without any major change in either `stats` or `ows`. Long term solution would be removing the thumbnail generation from `stats` and have it as an standalone after fact processing functionality.


odc/stats/io.py
```diff
@@ -80,6 +81,15 @@ def _sha1_digest(*write_results):
     return "".join(lines)
 
 
+@dask.delayed(name="xarray-to-list")
+def _xarray_to_list(image, dest_shape):
+    # apply the OWS styling, the return image must have red, green, blue and alpha.
+    display_pixels = []
+    for display_band in ['red', 'green', 'blue']:
```
alexgleith, line 88: The output will be RGB, but the input could be anything, right?

supermarkion (approved):
Looks good to me.

odc/stats/io.py
emmaai: not anything, but a few things, e.g. `dict`, `xr.Dataset` or `dask.Delayed`

odc/stats/io.py
```diff
@@ -242,13 +246,17 @@ def _ds_to_thumbnail_cog(self, ds: xr.Dataset, task: Task) -> List[Delayed]:
...
-                _log.error(f"Cannot parse OWS styling: {task.product.preview_image_ows_style}.")
+                image = self._apply_color_ramp(ds, task.product.preview_image_ows_style, task.time_range.start)
+            except AttributeError as e:
+                _log.error(f"{e} Cannot parse OWS styling: {task.product.preview_image_ows_style}.")
             except ImportError as e:
                 raise type(e)(str(e) + '. Please run python -m pip install "odc-stats[ows]" to setup environment to generate thumbnail.')
+            else:
+                display_pixels = _xarray_to_list(image, task.geobox.shape[0:2])
+                thumbnail_cog = self._get_thumbnail(display_pixels, input_geobox, odc_file_path)
+                thumbnail_cogs.append(thumbnail_cog)
```
emmaai, line 259: I got question for this actually, what was the purpose? from my understanding of the original code, if ows style failed, it wouldn't generate thumbnail but continue normally for the rest, then what's the point of raising `ImportError` then? Shouldn't we raise all or ignore all?

supermarkion: `ImportError` aims to display this message `run python -m pip install odc-stats[ows]`. Just in case some users may setup their odc-stats apps by `run python -m pip install odc-stats` but still try to use OWS features.

SpacemanPaul (approved):
This makes sense to me.  Nice clean PR.

PR #22, opened 2022-02-22
Add in CLI option to handle collection_category flag

alexgleith:
Simpler than I thought it would be.

Turns out we really shouldn't have implemented the `dataset_filter` option and just used this existing predicate logic...

### DIFF
--- a/odc/stats/_cli_save_tasks.py
+++ b/odc/stats/_cli_save_tasks.py
@@ -64,6 +64,11 @@ from itertools import groupby
     help="Only save datasets that pass `gqa_iterative_mean_xy <= gqa` test",
 )
 @click.option(
+    "--usgs-collection-category",
+    type=str,
+    help="Only save datasets that pass `collection_category == usgs_collection_category` test",
+)
+@click.option(
     "--dataset-filter",
     type=str,
     default=None,
@@ -85,6 +90,7 @@ def save_tasks(
     tiles=None,
     debug=False,
     gqa=None,
+    usgs_collection_category=None,
 ):
     """
     Prepare tasks for processing (query db).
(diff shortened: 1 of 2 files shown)


odc/stats/_cli_save_tasks.py
```diff
@@ -153,9 +159,18 @@ def on_message(msg):
     def gqa_predicate(ds):
         return ds.metadata.gqa_iterative_mean_xy <= gqa
 
+    def collection_category_predicate(ds):
+        if ds.type in ["ls5_sr", "ls7_sr", "ls8_sr", "ls9_sr"]:
+            return ds.metadata.collection_category == usgs_collection_category
+        else:
+            return True
```
SpacemanPaul, line 166: This implementation is weird to me.  Why isn't it just: 

`return ds.metadata.collection_category == usgs_collection_category`

SpacemanPaul (commented):
The current predicate implementation seems very odd to me.

Why are `lsn_sr` products handled separately, and why are they hard-coded?  Why does it default to True for non `lsn_sr` products?

odc/stats/_cli_save_tasks.py
alexgleith: Because not all products have `collection_category`, and so it filters out all the Sentinel-2 datasets, for example.

Our use case is to be able to run the command:

```
odc-stats save-tasks --grid africa_30 --temporal-range 2021-12--P1M --usgs-collection-category=T2 ls8_sr-ls9_sr-s2_l2a
```

SpacemanPaul: It just feels very specific.  I really don't like the hard-coded product names.

Can you add another CLI option, e.g.

```
odc-stats save-tasks --grid africa_30 --temporal-range 2021-12--P1M --usgs-collection-category=T2 --usgs-collection-products=ls8_sr-ls9_sr ls8_sr-ls9_sr-s2_l2a
```

cbur24 (approved):

SpacemanPaul (approved):

PR #26, opened 2022-03-01
small fixes on mangroves plugin

emmaai:
Fixes to get mangroves into production
- register the plugin
- fix the mistake accidentally pass `mangroves_extent` to the parent class
- warn the shape file not found
- remove the unwanted dimension from output


SpacemanPaul (approved):
I'm not real familiar with what the mangroves plugin was doing before, but these changes all make sense.

PR #27, opened 2022-03-22
change failure exit code

emmaai:
Require a non 0 exit code to mark the failure and trigger retry. It is not a good practice to exit everything with 0.
Scenario on k8s:
- `liveness probe` run every 5s
- within 5s interval stats failed and exit 0
- nobody knows which ones failed which one succeeded


jmettes (approved):

omad (approved):
That looks better.

PR #28, opened 2022-03-23
Produce all bands for tci percentile

emmaai:
- produce percentiles for the other two indices
- make output bands configurable, in case we change minds again
- make compute part code flow better


omad (approved):
lgtm

odc/stats/plugins/tcw_percentiles.py
```diff
@@ -42,23 +51,22 @@ def native_transform(self, xx):
...
-        tcw.attrs['nodata'] = NODATA
-        
-        xx = xx.drop_vars(xx.data_vars.keys())
-        xx['wet'] = tcw.astype(np.int16)
-        xx = keep_good_only(xx, ~bad, nodata=NODATA)
-        return xx
+        yy = xx.drop_vars(self.input_bands)
+        for m in self.output_bands:
+            yy[m] = sum(coeff * xx[band] for band, coeff in self.coefficients[m].items()).astype(np.int16)
+            yy[m].attrs = xx.blue.attrs
```
jmettes, line 61: Should there be other colours besides blue be considered for other two indices? (just spitballing here - I'm not familiar with the algorithm)

jmettes (approved):

odc/stats/plugins/tcw_percentiles.py
emmaai: the output was supposed to inherit the `attrs` from any input band, they're all the same other than `nodata`. So I just picked blue.

PR #29, opened 2022-03-25
refactor querying multiple products

emmaai:
Ref: https://github.com/opendatacube/odc-stats/commit/7dfc4168dc1f7ad307d94e039f712069a9b72d7d
problem: pull all the datasets of the products from the db even when the query was for a small area, e.g., a single tile. It caused:
1. significant delay for the db query when scaling the jobs, rendered granulising jobs invalid
2. blew up the cache/ephemeral storage unnecessarily 

It's an overhaul of `save-tasks`
- gather similar functionalities together
- better hierarchy of interfaces
- remove some unfavourable practices, e.g., unpack the iterator when not really in need
- disable some debug functions since a) too distracting b) the argument above

What's more:
- fix the bugs when the dataset size to small to train the dictionary


odc/stats/tasks.py
```diff
@@ -130,10 +129,23 @@ def __init__(
...
         return out_path(suffix, self._output)
 
+    def ds_align(self, dss: Iterable, product: DatasetType, pair_size: int):
+
+        def pair_dss(groups, pair_size):
+            for _, ds_group in groups:
+                ds_group = tuple(ds_group)
+                if len(ds_group) == pair_size:
+                    yield ds_group
+        paired_dss = groupby(dss, key=lambda ds: (ds.center_time, ds.metadata.region_code))
```
alexgleith: I'm pretty sure `center_time` always exists but `ds.metadata.region_code` is MetadataType specific.

Not sure a better approach, but it might be worth checking for the existence of a region_code and handling if it doesn't exist.

alexgleith (commented):
Looks fine to me. One note about the use of `region_code`, which is not guaranteed to exist. (That said it DOES for all DE Africa and DE Australia data, I think, so...)

Can you confirm the resulting merged products and multi-products are the same as existing ones?

I should look and see if there's test coverage...

odc/stats/tasks.py
```diff
@@ -130,10 +129,23 @@ def __init__(
...
+        paired_dss = pair_dss(paired_dss, pair_size)
+        map_fuse_func = lambda x: fuse_ds(*x, product=product)
+        dss = map(map_fuse_func, paired_dss)
+        return dss
+
```
SpacemanPaul, line 144: I find this bit confusing.  "Pair" implies 2 of something, but pair_size is a free parameter here.

Also the line:  `paired_dss = pair_dss(paired_dss, pair_size)` seems discard any groups whose size does not match pair_size (rather than trying to partition them down or aggregate them up into subgroups of exact size pair_size).  Is this the intention?

odc/stats/tasks.py
```diff
@@ -238,118 +234,95 @@ def msg_default(msg):
...
+            tasks = bin_annual(cells)
+        elif temporal_range is not None:
+            tasks = bin_generic(cells, [temporal_range])
+        else:
+            tasks = bin_annual(cells)
+        # Remove duplicate source uuids.
+        # Duplicates occur when queried datasets are captured around UTC midnight
+        # and around weekly boundary
+        tasks = {k: set(dss) for k, dss in tasks.items()}
+        tasks_uuid = {k: [ds.id for ds in dss] for k, dss in tasks.items()}
```
SpacemanPaul, line 319: Could you just use `list(dss)` in place of `[ds.id for ds in dss]`?

SpacemanPaul (changes requested):
Mostly looks good.  I think the "pair_dss" function either needs more work or a less confusing name.

odc/stats/tasks.py
emmaai: Dataset has many things, but here is to save only id of datasets in the task uuid. I guess it was for publishing to sqs. Though we don't do that anymore.

odc/stats/tasks.py
emmaai: I'll change the function name, it's to gather >= 2 datasets in group, hence the free pair_size. And yes that's the intention. If datasets missing from one of products, we won't process that 'group' altogether.

emmaai: done

odc/stats/tasks.py
emmaai: done

emmaai:
Find it a bit hard to test `save-tasks` here, better do this in `statistician` docker. I confirmed the datasets are the same with before the change.

alexgleith (approved):

SpacemanPaul (approved):

PR #32, opened 2022-03-29
Fix multi db support

alexgleith:
* This should fix the case where multiple DBs are being used in the one SQS queue

Closes #32

### DIFF
--- a/odc/stats/tasks.py
+++ b/odc/stats/tasks.py
@@ -421,7 +421,7 @@ class TaskReader:
             gs.crs, gs.tile_size, resolution=resolution, origin=gs.origin
         )
 
-    def init_from_sqs(self, cache: Union[str, DatasetCache]):
+    def init_from_sqs(self, local_db_path: str):
         """
         Adding the missing _grid, _gridspec, _gridspec and _all_tiles which skip for sqs task init.
         Upading the cfg which used placeholder filedb path for sqs task init.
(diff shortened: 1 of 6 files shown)


odc/stats/tasks.py
```diff
@@ -544,14 +543,22 @@ def stream_from_sqs(
...
-            else: # if it is the test_tiles.db, we load it from local
+            else:
+                # If it is the test_tiles.db, we load it from local
                 key = filedb
             local_cache_file = key.split("/")[-1]
-            if not os.path.isfile(local_cache_file):  # use the download filedb from S3 as the init context flag
+            # Use the downloaded filedb from S3 as the init context flag
+            if not os.path.isfile(local_cache_file):
                 self.init_from_sqs(filedb)
+            else:
```
supermarkion, line 551: My original fix plan is: 

No matter we have the db_cache file locally or not, we will always run `self.init_from_sqs(filedb)`. But we do the if-else check on this https://github.com/opendatacube/odc-stats/blob/develop/odc/stats/tasks.py#L383 to not download same db_cache file twice.

Of course, your version also works.

supermarkion (approved):
Looks all good to me.

odc/stats/tasks.py
alexgleith: I think mine is simpler... but yours updates those other fields, so, I think yours is safer.

I'll change this PR.

supermarkion (approved):
Looks very good to me!

supermarkion (approved):
All good. Cheers

PR #34, opened 2022-03-30
Add exception handling to task running

alexgleith:
Not sure if this is naive, but I have some tasks failing, and Statistician currently crashes, rather than marking the task as an error and continuing.

I _think_ this is the right place to catch exceptions and gracefully return an error notification.

### DIFF
--- a/odc/stats/proc.py
+++ b/odc/stats/proc.py
@@ -202,40 +202,44 @@ class TaskRunner:
                     yield TaskResult(task, path, skipped=True)
                     continue
 
-            _log.debug("Building Dask Graph")
-            ds = proc.reduce(proc.input_data(task.datasets, task.geobox))
+            try:
+                _log.debug("Building Dask Graph")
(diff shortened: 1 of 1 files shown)


supermarkion (approved):
Should be worth to have a try. Better than no try-catch.

emmaai (commented):
I have my doubt on this "catching everything then move on" approach. And I'm not sure about the context, if it's the only resort to fix the problem, then yes, if we could do more work to avoid crash, then we should.

alexgleith:
> I have my doubt on this "catching everything then move on" approach. And I'm not sure about the context, if it's the only resort to fix the problem, then yes, if we could do more work to avoid crash, then we should.

I agree that crashes shouldn't happen, but if they do, we have a choice of how we handle them. Currently, it's basically by aborting the whole job. What should happen is that Task should fail, be logged and the next Task can be tried.

This is especially important for streaming from SQS, which has built-in retries and failed task logging using the deadletter queue.

PR #35, opened 2022-03-31
Fix query freq

emmaai:
Had a look at the `odc-dcache` code, realised chopping the query into chunks too small is a bad idea
- take even longer when the query hit empty (it's bit funny but happened)
- doesn't speed up the query, and because of the reason above, it slows the summary product query significantly 5s->1m45s

Hence keep all the datasets query frequency to yearly, for the benefit of both bulky (multi years) and small (<= a year) queries.


SpacemanPaul (approved):
Good PR, but not as good as cake.

alexgleith (approved):
Looks fine, I don't know what the implications are though.

emmaai:
My understanding of chopping query with time frequency is to reduce the datasets size moved around the memory, hence reducing the load of database given the shared memory limit. In `odc-dcache` code, achieved by sending multi-queries to db. So the change might slow some datasets query, e.g., query all datasets in the database within a year.

emmaai:
My understanding of chopping query with time frequency is to reduce the datasets size moved around the memory, hence reducing the load of database given the shared memory limit. In `odc-dcache` code, achieved by sending multi-queries to db. So the change might slow some datasets query, e.g., query all datasets in the database within a year.

PR #36, opened 2022-04-05
Add geometry filter to task publishing

alexgleith:
Context is that for DE Africa we have a tool that does this, but it's in two places, and isn't implemented particularly well. So I thought I'd implement it somewhat decently over here so that I can use it and so that others can also.

It handles loading some GeoJSON and filtering tiles to that region before pushing them to SQS.


emmaai (approved):
LGTM

PR #38, opened 2022-04-08
Fix typo in base rgba method.

SpacemanPaul:
Typo in base plugin rgba method - reported by Andres Sutton on ODC Slack.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -18,6 +18,8 @@ zip_safe = false
 packages = find_namespace:
 python_requires = >=3.8
 tests_require = pytest
+build_requires = 
+    packaging
 install_requires =
     botocore
     click>=8.0.0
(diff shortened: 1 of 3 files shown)


setup.cfg
```diff
@@ -18,6 +18,8 @@ zip_safe = false
 packages = find_namespace:
 python_requires = >=3.8
 tests_require = pytest
+setup_requires =
```
Kirill888: I don't think this is needed here, packaging is used on githubactions only, to patch version numbers before creating wheels/sources, not when using those to install the package.

SpacemanPaul: Yeah I wasn't sure if that was still needed.  If you think it isn't I'll remove it again.

Kirill888 (approved):

emmaai (approved):
💯

PR #42, opened 2022-05-23
keep file name in url

emmaai:
To match the behaviour of `urllib.parse.urljoin`. Re: the assets paths missing one level in stac json

### DIFF
--- a/odc/stats/io.py
+++ b/odc/stats/io.py
@@ -293,7 +293,7 @@ class S3COGSink:
         _u = urlparse(stac_file_path)
 
         if _u.scheme == "s3":
-            dataset_location = f"{_u.scheme}://{_u.netloc}/{'/'.join(_u.path.split('/')[:-1])}"
+            dataset_location = f"{_u.scheme}://{_u.netloc}/{_u.path}"
         else:
             dataset_location = str(Path(_u.path).parent)
 


odc/stats/io.py
```diff
@@ -293,7 +293,7 @@ def get_eo3_stac_meta(self, task: Task, meta: DatasetDoc, stac_file_path: str, o
         _u = urlparse(stac_file_path)
 
         if _u.scheme == "s3":
-            dataset_location = f"{_u.scheme}://{_u.netloc}/{'/'.join(_u.path.split('/')[:-1])}"
+            dataset_location = f"{_u.scheme}://{_u.netloc}/{_u.path}"
```
omad, line 296: I'm not sure what the implications of this are, it's a not insignificant change, from stripping the last element, to not stripping the last element.

emmaai: The only usage of this parameter is in this function of making up the assets path
https://github.com/opendatacube/eo-datasets/blob/09dfd347dd7c47b5d05a04c9b70291cc9d1971c5/eodatasets3/stac.py#L308
and 
https://github.com/opendatacube/eo-datasets/blob/09dfd347dd7c47b5d05a04c9b70291cc9d1971c5/eodatasets3/stac.py#L339
where it calls the function  from `datacube.core`
https://github.com/opendatacube/eo-datasets/blob/09dfd347dd7c47b5d05a04c9b70291cc9d1971c5/eodatasets3/stac.py#L400
which calls `urllib.parse.urljoin`. So it doesn't affect anything else but the assets path.

emmaai: and from this line https://github.com/opendatacube/eo-datasets/blob/09dfd347dd7c47b5d05a04c9b70291cc9d1971c5/eodatasets3/stac.py#L251, the file path is not stripped if the `dataset_location` comes from the `metadata`.

omad: Awesome.

Sorry, I completely missed your replies Emma. Lets merge this in. :D

omad (approved):

PR #44, opened 2022-06-01
Change how the plugin config is overwritten and wo summary load data

emmaai:
re: plugin config
-------
Current behaviour:
- plugin config from config yaml got totally overwritten by cli options

After change:
- only the same entry got overwritten, keep the different entries untouched and added from cli options

Reasoning:
Some configs can be pre-defined, some of them could be generated dynamically in the running environment, e.g., numerical parameters vs artefacts path and/or prefix. We wanna keep pre-defined terms in the config file as much as possible. They shouldn't be overwritten had it not been intended. Also we wanna append all the dynamic configs when it happens.

re: wo all-time summary data loading
------
data were loaded with a function from `odc.stac`
1. the function is obsoleted, `odc.stac.load` is loading specifically with `pystac.item` now
2. it shouldn't have been such, as the super class provides default data loading method

### DIFF
--- a/odc/stats/_cli_run.py
+++ b/odc/stats/_cli_run.py
@@ -140,7 +140,7 @@ def run(
 
     _cfg.update(cfg_from_cli)
     if plugin_config is not None:
-        _cfg["plugin_config"] = plugin_config
+        _cfg["plugin_config"].update(plugin_config)
 
     if resampling is not None and len(resampling) > 0:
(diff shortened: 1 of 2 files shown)


erin-telfer:
Looks fine, will need review from someone who is more technical than me.

tebadi (approved):
LGTM!

PR #48, opened 2022-06-07
Separate cloud and shadow masks in geomedian.

tebadi:


odc/stats/plugins/gm.py
```diff
@@ -78,18 +78,25 @@ def native_transform(self, xx: xr.Dataset) -> xr.Dataset:
...
-            self._mask_band,
-            geobox,
-            categories=self.cloud_classes,
-            filters=self.filters,
-            groupby=self.group_by,
-            resampling=self.resampling,
-            chunks={},
-        )
+        cloud_shadow_mask = None
+        for cloud_class, filter in zip(self.cloud_classes, self.filters):
```
emmaai: I don't think it'll be working without changing the interface of the class.
-  `load_enum_filtered` requires a list of `str` or `int` as categories, and a list of `tuple` as filters. After this line, the input will be `str` or `int`, and `tuple` respectively. 

- In this change, it's assuming all the `cloud_classes` will be dealt separately. It was not `load_enum_filtered` intended for hence it's not what this input  intended for. We might want to get this very clear by separating two scenarios, a) smash all `cloud_classes` together to have one boolean, or b) keep them separate to have multiple booleans.

emmaai (changes requested):

odc/stats/plugins/gm.py
tebadi: The assumption here is that the input filter looks something like this:
`[[["opening", 0], ["dilation", 5]], [["opening", 0], ["dilation", 10]]]` 

This will be a change in the input.  I don't think we need to keep the single filter support at least for the summary products.   I'm happy to add the support if we think it's needed.

tebadi: I still need to update the type annotation for filters.

emmaai: so, if that's the case the `cloud_classes` will have the form of `[[a], [b]]` as well? it'll support the single filter by input being `[[a, b]]`.

tebadi: Alternatively I can add a separate input argument for separate filters to the gm plugin.  something that accepts a dictionary where keys are cloud classes and values are filters:
```
{'cloud': [['opening', 0], ['dilation', 5]],
 'shadow': [['opening', 0], ['dilation', 5]]}
```

emmaai: I like this one better actually. Easier to format in the config yaml as well.

odc/stats/plugins/gm.py
```diff
@@ -23,6 +23,7 @@ def __init__(
         cloud_classes: Tuple[str, ...],
         nodata_classes: Optional[Tuple[str, ...]] = None,
         filters: Optional[Iterable[Tuple[str, int]]] = None,
+        cloud_filters: Optional[Dict[str, Iterable[Tuple[str, int]]]]=None,
```
emmaai: Think we can have `Dict[Tuple[str], ...]` here, so that we will get rid of `cloud_classes` and `filters`? Also there is no need of line `80-91` or constructing a `tuple` in line `99` in this way.

tebadi: Hmm no, we have currently Dict[str, ..].  Dict[Tuple[str], ...] comes at the cost of ugly input:
 ('cloud', ): [['opening', 0], ['dilation', 5]],
 ('shadow', ): [['opening', 0], ['dilation', 5]]

I had kept the filters and cloud_classes for backwards compatibility but removed it since even if we need this feature, we can still achieve it by reformatting the input.

emmaai: Agreed, they’re indeed ugly once having values filled…

odc/stats/plugins/gm.py
```diff
@@ -20,10 +20,9 @@ def __init__(
         self,
         bands: Tuple[str, ...],
         mask_band: str,
-        cloud_classes: Tuple[str, ...],
         nodata_classes: Optional[Tuple[str, ...]] = None,
-        filters: Optional[Iterable[Tuple[str, int]]] = None,
-        basis_band=None,
+        cloud_filters: Optional[Dict[str, Iterable[Tuple[str, int]]]] = None,
```
emmaai: In this case `cloud_filters` is not `optional` anymore, only the `value` of `Dict` (dilation strategy) is. Otherwise, if there is no `dilation`, there won't be a cloud/shadow mask at all.

tebadi:
This is addressed in another PR as part of geomedian refactoring.

PR #50, opened 2022-06-21
Geomedian refactor

tebadi:
This PR refactors geomedian plugin:
- It adds a check for contiguity in data
- all the masking including nodata, cloud buffering and the contiguity occur prior to reprojection.
- It keeps good data vs previous implementation that kept bad data (no data and cloud masked together)
- Removed two levels of data-load - previously cloud data was loaded using `load_enum_filtered` generate a cloud mask which was then applied to remove bad data after a second level of load

### DIFF
--- a/odc/stats/plugins/gm.py
+++ b/odc/stats/plugins/gm.py
@@ -1,13 +1,13 @@
 """
 Geomedian
 """
-from typing import Optional, Mapping, Sequence, Tuple, Iterable
+from typing import Optional, Mapping, Sequence, Tuple, Iterable, Dict
 import xarray as xr
 from datacube.model import Dataset
 from datacube.utils.geometry import GeoBox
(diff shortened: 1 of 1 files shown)


odc/stats/plugins/gm.py
```diff
@@ -67,29 +62,28 @@ def native_transform(self, xx: xr.Dataset) -> xr.Dataset:
...
-            datasets,
-            self._mask_band,
-            geobox,
-            categories=self.cloud_classes,
-            filters=self.filters,
-            groupby=self.group_by,
-            resampling=self.resampling,
-            chunks={},
-        )
+    def input_data(
```
emmaai: we don't need this function here anymore

emmaai (changes requested):

emmaai (approved):
Thanks!

PR #51, opened 2022-06-21
Ensure GeoMAD band names match the band name of C3 Landsat data.

tebadi:
Currently the GeoMAD SR band names differ from Collection 3 NBART product.  The issue was reported by Robbie :

image

where left is C3 Landsat band names and right is the current GeoMAD. 

This fix is addressing the inconsistency  issue.

### DIFF
--- a/odc/stats/plugins/gm.py
+++ b/odc/stats/plugins/gm.py
@@ -53,12 +53,6 @@ class StatsGM(StatsPluginInterface):
                 "emad",
                 "bcmad",
                 "count",
-                "red",
-                "green",
-                "blue",
-                "nir",
-                "swir1",
-                "swir2",
             )
         )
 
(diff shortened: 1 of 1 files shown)


robbibt (approved):
Looks great, thanks @tebadi! I imagine the band aliases themselves can be updated somewhere downstream/not in `stats`?
image (in case we wanted to add the old names as aliases to match C3 Landsat)

tebadi:
> 

Yes, that will be changed in the product definitions Robbi.

emmaai:
I don't think we need do this? We can just add aliases in the product definition as ARD products have been doing.

tebadi:
> I don't think we need do this? We can just add aliases in the product definition as ARD products have been doing.

This is fixing it at the source instead of fixing it downstream.

emmaai (commented):
If that’s the argument, then the input bands name should be fixed rather than hard mapping here. It was to take input band name as the output band name. The reason that we have seemingly different band names than what you preferred is because it loaded the bands by one of many aliases provided by ard. If it only loads the other aliases you like, it’ll give you the preferred band names.

Besides, this hard mapping is so fragile,  slightly change on inputs bands name will render the whole thing broken, the same example that the input bands aliases change.

tebadi:
> If that’s the argument, then the input bands name should be fixed rather than hard mapping here. It was to take input band name as the output band name. The reason that we have seemingly different band names than what you preferred is because it loaded the bands by one of many aliases provided by ard. If it only loads the other aliases you like, it’ll give you the preferred band names.

I'd like fixing the band names to the preferred/output band names as opposed to adding alias here.  It's more readable and achieves what we need for consistency between ARD and GeoMAD spectral bands.   Done.

PR #52, opened 2022-06-22
load all configs from one single yaml

emmaai:
Aimed to gather all the configs to produce a summary product in one single yaml, to 

- improve the consistency from run to run, and between sandbox/devbox and k8s
- reduce the barrier for automation
- be easier for maintenance and not collecting bits and pieces from different repos just for getting a right config

Changes

- `save-tasks` and `run` loading configs from the same yaml
- sanity checks on what configs are allowed by the functionalities
- add `.pre-commit-config.yaml` and `black` in GitHub workflow
- all files got formatted by `black`

It should be back-compatible, no default behaviour change is intended or expected.

### DIFF
--- a/odc/stats/_cli_save_tasks.py
+++ b/odc/stats/_cli_save_tasks.py
@@ -107,7 +107,7 @@ def save_tasks(
       - output product config
 
     """
-    setup_logging(0)
+    setup_logging()
 
     import logging
     from datacube import Datacube
@@ -122,10 +122,8 @@ def save_tasks(
     if config is None:
         config = {}
 
-    _log.debug(f"read in config : {config}")
-    _cfg = {k: config.get(k) for k in CONFIG_ITEMS}
+    _cfg = {k: config.get(k) for k in CONFIG_ITEMS if config.get(k) is not None}
 
-    _log.debug(f"read in config : {_cfg}")
     cfg_from_cli = {
         k: v
         for k, v in dict(
@@ -141,6 +139,7 @@ def save_tasks(
 
     _log.info(f"Config overrides: {cfg_from_cli}")
     _cfg.update(cfg_from_cli)
+    _log.info(f"Using config: {_cfg}")
 
     gqa = _cfg.pop("gqa", None)
     products = _cfg.pop("products", None)
--- a/odc/stats/_cli_save_tasks.py
+++ b/odc/stats/_cli_save_tasks.py
@@ -2,7 +2,17 @@ import json
 
 import click
 import sys
-from ._cli_common import main, click_range2d
+from ._cli_common import main, click_range2d, click_yaml_cfg, setup_logging
+
+CONFIG_ITEMS = [
+    "grid",
+    "frequency",
+    "complevel",
+    "overwrite",
+    "gqa",
+    "products",
+    "dataset_filter",
+]
 
 
 @main.command("save-tasks")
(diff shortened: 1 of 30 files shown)


odc/stats/plugins/tcw_percentiles.py
```diff
@@ -23,16 +23,43 @@ class StatsTCWPC(StatsPluginInterface):
...
+        input_bands: Sequence[str] = [
+            "blue",
+            "green",
+            "red",
+            "nir",
+            "swir1",
+            "swir2",
+            "fmask",
+            "nbart_contiguity",
+        ],
```
tebadi, line 60: I think we should change the spectral band names to have the "nbart" prefix to match the ARD and GeoMAD bands.

tebadi (commented):
I only checked the stats code and that looks good to me.  I noticed one inconsistency in band names for tcw.

SpacemanPaul (approved):
This is a daunting and confusing PR to review because it consists mostly of formatting and whitespace changes.

The bit that are actual code changes look reasonable to me, but I can't guarantee that I haven't missed something.

odc/stats/plugins/tcw_percentiles.py
emmaai: `tc-percentiles` won't have any input band as output, unlike GeoMAD, and this line stated the output bands naming explicitly
https://github.com/opendatacube/odc-stats/blob/5479f91ce3fc81d013e1321c75f2a33ea81a54bc/odc/stats/plugins/tcw_percentiles.py#L34

odc/stats/_cli_save_tasks.py
```diff
@@ -99,18 +107,70 @@ def save_tasks(
...
+
+    cfg_from_cli = {
+        k: v
+        for k, v in dict(
+            grid=grid,
+            frequency=frequency,
+            gqa=gqa,
+            products=products,
+            complevel=complevel,
+            dataset_filter=dataset_filter,
```
omad, line 135: `overwrite` should be a part of this list.

odc/stats/_cli_save_tasks.py
```diff
@@ -99,18 +107,70 @@ def save_tasks(
...
+    gqa = _cfg.pop("gqa", None)
+    products = _cfg.pop("products", None)
+    dataset_filter = _cfg.pop("dataset_filter", None)
+
+    if products is None:
+        print("Input products has to be specified", file=sys.stderr)
+        sys.exit(1)
+
+    if _cfg.get("grid") is None:
+        print(
```
omad, line 153: Should print errors to `sys.stderr` like above. :)

I'm sure we don't do it everywhere, but I saw this one. :)

odc/stats/_cli_save_tasks.py
```diff
@@ -99,18 +107,70 @@ def save_tasks(
...
+        if _cfg.get("frequency") not in (
+            "annual",
+            "annual-fy",
+            "semiannual",
+            "seasonal",
+            "nov-mar",
+            "apr-oct",
+            "all",
+        ):
+            print(
```
omad, line 169: Errors to `stderr`, as above.

omad (approved):
Looks like a great change Emma. Thanks

odc/stats/plugins/tcw_percentiles.py
tebadi: What plugin generates this: 
image

emmaai: this plugin.

tebadi: This looks good.

tebadi (approved):

odc/stats/_cli_save_tasks.py
emmaai: note to myself, fix this in wofs pr

odc/stats/_cli_save_tasks.py
emmaai: note to myself, fix this in wofs pr

odc/stats/_cli_save_tasks.py
emmaai: note to myself, `overwrite` is a bit odd atm, seems like it appends the old cache, rather than "overwrite". get on it later when I'm sure what's supposed to happen.

PR #55, opened 2022-06-27
Apply cloud buffering to Tasseled Cap percentile. 

tebadi:


.pre-commit-config.yaml
```diff
@@ -25,6 +25,6 @@ repos:
     hooks:
       - id: flake8
   - repo: https://github.com/psf/black
-    rev: 20.8b1
+    rev: 22.3.0
```
tebadi, line 28: the previous version of black was throwing the following exception:
```
Traceback (most recent call last):
  File "/Users/toktamebadi/.cache/pre-commit/repo4r1vocgg/py_env-python3/bin/black", line 8, in <module>
    sys.exit(patched_main())
  File "/Users/toktamebadi/.cache/pre-commit/repo4r1vocgg/py_env-python3/lib/python3.8/site-packages/black/__init__.py", line 6606, in patched_main
    patch_click()
  File "/Users/toktamebadi/.cache/pre-commit/repo4r1vocgg/py_env-python3/lib/python3.8/site-packages/black/__init__.py", line 6595, in patch_click
    from click import _unicodefun  # type: ignore
ImportError: cannot import name '_unicodefun' from 'click' (/Users/toktamebadi/.cache/pre-commit/repo4r1vocgg/py_env-python3/lib/python3.8/site-packages/click/__init__.py)
```

odc/stats/plugins/tcw_percentiles.py
```diff
@@ -22,48 +27,77 @@ class StatsTCWPC(StatsPluginInterface):
...
+                "nbart_contiguity",
+            ]
+            if input_bands is None
+            else input_bands
+        )
+
+        super().__init__(input_bands=self.input_bands, **kwargs)
+        self.output_bands = (
+            ["wet", "bright", "green"] if output_bands is None else output_bands
+        )
```
tebadi, line 103: All of this formatting change was done to keep black happy.

emmaai: thanks!

odc/stats/plugins/tcw_percentiles.py
```diff
@@ -76,10 +110,15 @@ def native_transform(self, xx):
...
-        bad |= xx["nbart_contiguity"] == 0  # or the nbart contiguity bit is 0
+        mask = xx["fmask"]
+        nodata = enum_to_bool(mask, ("nodata",))
+        non_contiguent = xx["nbart_contiguity"] == 0
+        bad = nodata | non_contiguent
+        # Now exclude cloud and cloud shadow (including buffered pixels)
+        for cloud_class, c_filter in self.cloud_filters.items():
+            cloud_mask = enum_to_bool(mask, (cloud_class,))
+            cloud_mask_buffered = mask_cleanup(cloud_mask, mask_filters=c_filter)
+            bad = cloud_mask_buffered | bad
```
emmaai: Can we keep the fall-back behaviour here where there is no dilation? for the sake of sanity check/test and backward compatibility.

emmaai (changes requested):

odc/stats/plugins/tcw_percentiles.py
tebadi: Added support for no dilation and added a log message so if done unintentionally we can see it in the log.

odc/stats/plugins/tcw_percentiles.py
```diff
@@ -76,10 +112,18 @@ def native_transform(self, xx):
...
+        non_contiguent = xx["nbart_contiguity"] == 0
+        bad = nodata | non_contiguent
+        # Now exclude cloud and cloud shadow (including buffered pixels)
+        if self.cloud_filters is not None:
+            for cloud_class, c_filter in self.cloud_filters.items():
+                cloud_mask = enum_to_bool(mask, (cloud_class,))
+                cloud_mask_buffered = mask_cleanup(cloud_mask, mask_filters=c_filter)
+                bad = cloud_mask_buffered | bad
+        else:
+            _log.info("There is no cloud/shadow buffering.")
```
emmaai, line 126: It won't do cloud/shadow mask at all if there is no dilation. We'll still need cloud/shadow mask as what they're if it's the case.

tebadi: Oh yeah, good catch.  I had changed the default behaviour!  Fixed :)

emmaai: `self.cloud_filters` won't be `None` since it was set `{}` at line 39

emmaai: I'll approve it. Functionality wise, it's ok. Coding wise, it's slightly messy and not ideal for dask graphs.

Note to myself: do some refactor when/if any change is required.

emmaai (approved):

PR #56, opened 2022-06-30
wofs and fc percentile plugin

emmaai:
Changes
- add cloud/shadow dilation to wofs and fc percentile plugin.
- add unit test for wofs plugin.
- apply mask before reprojection in wofs plugin.
- fix a bug in fc percentile, where `dry` and `wet` should mutually exclusive after fusing by keeping `dry`.
- fix a bug in wofs, where `non-contiguity` should be treated constantly as `non-clear` in masking, fusing and summarising. 
- moderate refactor of fc percentile to remove too many local variables introduced by DEAfrica workflow.
- major refactor of wofs to reflect the logic applied in fusing and counting/summarising.

Note:
There is no test on cloud/shadow dilation since it'll require some larger datasets. It'll be done in either `stats-docker` workflow or here after I can copy/paste Toktam's solution on GeoMAD test.

My apologies that this is a slightly large PR, due to the fact that I was switching between these two plugins.

### DIFF
--- a/odc/stats/plugins/fc_percentiles.py
+++ b/odc/stats/plugins/fc_percentiles.py
@@ -3,12 +3,12 @@ Fractional Cover Percentiles
 """
 from functools import partial
 from itertools import product
-from typing import Optional, Tuple
+from typing import Tuple, Dict, Iterable, Optional
 
 import numpy as np
 import xarray as xr
 from odc.algo import keep_good_only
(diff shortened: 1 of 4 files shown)


SpacemanPaul (approved):
I don't totally follow exactly what's happening, but the code looks solid and the bits I do follow make eminent sense.

tebadi (commented):
Let's do a live review - I don't understand a few things.

tebadi (approved):
Looks good!

PR #57, opened 2022-07-01
Ls tests

tebadi:


odc/stats/plugins/gm.py
```diff
@@ -75,11 +83,15 @@ def native_transform(self, xx: xr.Dataset) -> xr.Dataset:
...
-        for cloud_class, filter in self.cloud_filters.items():
-            cloud_mask = enum_to_bool(mask, (cloud_class,))
-            cloud_mask_buffered = mask_cleanup(cloud_mask, mask_filters=filter)
-            bad = cloud_mask_buffered | bad
+        if self.cloud_filters is not None:
+            for cloud_class, c_filter in self.cloud_filters.items():
+                cloud_mask = enum_to_bool(mask, (cloud_class,))
+                cloud_mask_buffered = mask_cleanup(cloud_mask, mask_filters=c_filter)
+                bad = cloud_mask_buffered | bad
+        else:
```
tebadi, line 91: Here I've changed the default behaviour which is to exclude cloud and shadow pixels.  This is the default behaviour in TCW.

tebadi, line 91: Here I've changed the default behaviour which is to exclude cloud and shadow pixels. This is the default behaviour in TCW.  @emmaai Could you confirm this is the default behaviour in wofs and FC.

emmaai (approved):
LGTM

PR #59, opened 2022-07-05
Added unit tests for Tasseled Cap percentile.

tebadi:


tests/test_tcp.py
```diff
@@ -0,0 +1,149 @@
...
+    result = xx.compute()
+    nodata = -9999
+    expected_results = np.array(
+        [
+            [[-150, -33], [-11, nodata]],
+            [[nodata, nodata], [nodata, nodata]],
+            [[-14, 0], [nodata, nodata]],
+        ]
+    )
+    assert (expected_results == result.wet.data).all()
```
emmaai, line 149: it doesn't test the fusing but only the loading and masking.

emmaai (changes requested):
Could you also add the test for fusing and computing, e.g., like fc percentiles and wofs? Thanks.

PR #60, opened 2022-07-27
Fix issues in plugins and version update

emmaai:
- Partially rolled back wofs for the sake of consistency, kept data type as boolean, turned out our implementation of interpolating boolean as numerical values then reproject with `bilinear` not equivalent to `nearest`. 
- removed `import gdal` and hence rewrote mangroves mask part, as `rasterio` doc said "they're mutually incompatible" (https://rasterio.readthedocs.io/en/latest/topics/switch.html).
- update all plugin versions to better track the change
- further cleaned the cli interface and addressed some issues left from the last pr
- pin `xarray==2022.3.0` due to the groupby bug on multi-indexed `DataSet`, an issue was raised here https://github.com/pydata/xarray/issues/6836  There are some other problems, but noway/need to investigate atm.
- `pylint` local, the precommit rev will be deprecated soon and more checking available locally, such as if package can be imported
- update stats version for release


SpacemanPaul (approved):
Looks solid to me.

.pre-commit-config.yaml
```diff
@@ -16,10 +17,18 @@ repos:
       - id: requirements-txt-fixer
       - id: check-added-large-files
       - id: check-merge-conflict
-  - repo: https://github.com/pre-commit/mirrors-pylint
-    rev: 'v3.0.0a5'  # Use the sha / tag you want to point at
+  - repo: local
```
omad: ooooh :( :(

What's wrong with merely updating this to a new sha?

I don't have pylint or most other dev tools installed globally since there's too many different versions available between different repos.

tebadi (approved):
Looks good!

.pre-commit-config.yaml
emmaai: > I don't have pylint or most other dev tools installed globally

that's a good point...didn't think about it. I'll roll it back to remote repo. One can do local `pylint` locally if they need more thorough check

Issue #40, opened 2022-05-02
SHA1Sum files have invalid format for use with CLI  `sha1sum`


omad:
The `.sha1` checksum files generated by `odc-stats` are rejected by both the GNU and Busybox versions of `sha1sum -c`.

These tools expect the filename and checksum to be separate by one or two space characters, whereas `odc-stats` uses a tab (`\t`) character.

```
$ ls
ga_ls_tc_pc_cyear_3_x08y25_2015--P1Y_final.odc-metadata.yaml  ga_ls_tc_pc_cyear_3_x08y25_2015--P1Y_final_green_pc_10.tif
ga_ls_tc_pc_cyear_3_x08y25_2015--P1Y_final.proc-info.yaml     ga_ls_tc_pc_cyear_3_x08y25_2015--P1Y_final_green_pc_50.tif
ga_ls_tc_pc_cyear_3_x08y25_2015--P1Y_final.sha1               ga_ls_tc_pc_cyear_3_x08y25_2015--P1Y_final_green_pc_90.tif
ga_ls_tc_pc_cyear_3_x08y25_2015--P1Y_final.stac-item.json     ga_ls_tc_pc_cyear_3_x08y25_2015--P1Y_final_thumbnail.jpg
ga_ls_tc_pc_cyear_3_x08y25_2015--P1Y_final_bright_pc_10.tif   ga_ls_tc_pc_cyear_3_x08y25_2015--P1Y_final_wet_pc_10.tif
ga_ls_tc_pc_cyear_3_x08y25_2015--P1Y_final_bright_pc_50.tif   ga_ls_tc_pc_cyear_3_x08y25_2015--P1Y_final_wet_pc_50.tif
ga_ls_tc_pc_cyear_3_x08y25_2015--P1Y_final_bright_pc_90.tif   ga_ls_tc_pc_cyear_3_x08y25_2015--P1Y_final_wet_pc_90.tif

$ sha1sum -cw ga_ls_tc_pc_cyear_3_x08y25_2015--P1Y_final.sha1
sha1sum: invalid format
sha1sum: invalid format
sha1sum: invalid format
sha1sum: invalid format
sha1sum: invalid format
sha1sum: invalid format
sha1sum: invalid format
sha1sum: invalid format
sha1sum: invalid format
sha1sum: invalid format
sha1sum: invalid format
sha1sum: invalid format
sha1sum: invalid format
sha1sum: WARNING: 13 of 13 computed checksums did NOT match

[...]

PR #62, opened 2022-08-10
add integration test and data integrity test

emmaai:
- fix sha1 file format https://github.com/opendatacube/odc-stats/issues/40
- fix geomad output band names, tolerate sentinel 2 where `non-contiguity` band does not exist
- add integration test and check data integrity
- change GitHub workflow in testing
- cleanup to please `pylint`


odc/stats/io.py
```diff
@@ -178,14 +179,15 @@ def verify_s3_credentials(self, test_uri: Optional[str] = None) -> bool:
         assert rr.path == test_uri
         return rr.error is None
 
+    # pylint: disable=invalid-name
```
SpacemanPaul, line 182: I think I'm with pylint on this.  The function argument should be called `content_type`, not `ContentType`.

SpacemanPaul (approved):
Mostly looks good to me - not as daunting as it first appeared.

I'd like to see all function arguments named according to standard Python conventions (see comment), but otherwise happy to approve.

odc/stats/io.py
emmaai: I'm with pylint too. Only not sure if this one is special on purpose.

.github/workflows/main.yml
```diff
@@ -57,112 +57,79 @@ jobs:
...
         run: |
-          echo "/tmp/test_env/bin" >> $GITHUB_PATH
+          apt update
+          DEBIAN_FRONTEND=noninteractive apt-get install -y \
+          --fix-missing --no-install-recommends \
+          nodejs \
+          zstd
+          python -m pip install --upgrade --no-cache pip
+          pip install --upgrade --no-cache -r \
+          tests/requirements.txt
```
tebadi, line 89: This line will update datacube-core version which could over time become different from the one used in the stats.

.github/workflows/main.yml
```diff
@@ -172,71 +139,3 @@ jobs:
         with:
```
tebadi, line 139: Apparently codecov/codecov-action@v1 s deprecated - see https://github.com/marketplace/actions/codecov

codecov/codecov-action@v3 is the latest.

.github/workflows/main.yml
emmaai: it is meant to be. It's a regression test since we always release stats first and then update the docker image. Hence it's to test the results will remain the same with newer version of libs compared to the ones built in the current docker image. Then we'll be confident (sort of) that the new docker image will be working as expected. The same test will be integrated into the docker image build workflow as well.

.github/workflows/main.yml
emmaai: note to myself, fix it in next pr.

.github/workflows/main.yml
tebadi: It makes sense, I didn't see the forward thinking.

tebadi (approved):

odc/stats/io.py
```diff
@@ -81,7 +82,7 @@ def _sha1_digest(*write_results):
         if wr.error is not None:
             raise IOError(f"Failed to write for: {wr.path}")
         file = wr.path.split("/")[-1]
-        lines.append(f"{wr.sha1}\t{file}\n")
+        lines.append(f"{wr.sha1}  {file}\n")
```
jmettes, line 85: Yay

PR #63, opened 2022-09-01
support explicit transform path

emmaai:
To address the issue caused by multiple transform options available in `proj >= 8.0`. After the change, an explicit path chosen by the authority code will be passed to reproject (related pr https://github.com/opendatacube/odc-tools/pull/500). Currently the transform path with authority code `epsg:9688` is default with gdal. The purpose of supporting an explicit transform path is to maintain the data consistency and integrity, as well as being aware and alerted of the similar issue in the future. 

We have decided there is no need to reprocess the data with the old transform path (code `epsg:1150`), where GDA94 and WGS84 is deemed equivalent while with `epsg:9688` it is not. More infos and discussions were in the emails. Hence all the data of summary products will be reprojected with the current transform path `epsg:9688` thereafter. They'll be specified in the relative config files for production.

### DIFF
--- a/odc/stats/plugins/_base.py
+++ b/odc/stats/plugins/_base.py
@@ -1,5 +1,5 @@
 from abc import ABC, abstractmethod
-from typing import Callable, Mapping, Optional, Sequence, Tuple
+from typing import Mapping, Optional, Sequence, Tuple
 
 import xarray as xr
 from datacube.model import Dataset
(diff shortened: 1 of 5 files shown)


odc/stats/plugins/_base.py
```diff
@@ -18,20 +18,24 @@ class StatsPluginInterface(ABC):
     def __init__(
         self,
         resampling: str = "bilinear",
-        input_bands: Sequence[str] = [],
-        chunks: Mapping[str, int] = {"y": -1, "x": -1},
+        input_bands: Sequence[str] = None,
+        chunks: Mapping[str, int] = None,
```
SpacemanPaul: These type hints are wrong.

Should be `Optional[Sequence[str]]` and `Optional[Mapping[str, int]]` respectively if they can take `None`.

odc/stats/plugins/_base.py
```diff
@@ -18,20 +18,24 @@ class StatsPluginInterface(ABC):
...
         rgb_bands: Optional[Sequence[str]] = None,
         rgb_clamp: Tuple[float, float] = (1.0, 3_000.0),
+        transform_code: Optional[str] = None,
+        area_of_interest: Optional[Sequence[float]] = None,
     ):
         self.resampling = resampling
-        self.input_bands = input_bands
-        self.chunks = chunks
+        self.input_bands = input_bands if input_bands is not None else []
+        self.chunks = chunks if chunks is not None else {"y": -1, "x": -1}
```
SpacemanPaul, line 32: Even if you're doing this. :)

SpacemanPaul (commented):
Just one pedantic typehint comment.  Looks good.

tests/requirements.txt
```diff
@@ -16,6 +17,7 @@ pytest
 pytest-cov
 pytest-httpserver
 pytest-timeout
+rasterio>=1.3.2
```
SpacemanPaul, line 20: Might need to pin xarray too as most recent release is buggy.

Core has been updated to have   'xarray>=0.9,!=2022.6.0' but that's not released yet, so a build off this requirements file will most likely result in installing the buggy xarray-22.6.0

tebadi (approved):
LGTM. Approved pending addressing minor issues that Paul has raised.

tests/requirements.txt
emmaai: `xarray` is pinned in both the `setup.cfg` and the docker image that the integration test used. This file is to "patch" the older production docker image for a regression test.

odc/stats/plugins/_base.py
emmaai: fixed in a new commit to the pr

PR #67, opened 2022-10-14
Dataset maturity

Ariana-B:
Output metadata takes dataset maturity from dataset if it is not specified in the config.

Testing in dev sandbox with geomedian gives expected results.


SpacemanPaul (approved):

Ariana-B:
@omad should this be merged or is it no longer relevant/necessary?

Issue #65, opened 2022-09-27
.proc-info.yaml odc-tools reference pointing to old odc-tools repo


jmettes:
Looks like process metadata yaml is still pointing to odc-tools

```
$ cat ga_ls_fc_pc_fyear_3_x25y28_2010-07-01--P1Y_final.proc-info.yaml
...
software_versions:
- name: eodatasets3
  url: https://github.com/GeoscienceAustralia/eo-datasets
  version: 0.28.1
- name: odc-stats
  url: https://github.com/opendatacube/odc-tools        <-- should be odc-stats
  version: 1.0.26
- name: ga_fc_percentiles
  url: https://github.com/opendatacube/odc-tools
  version: 0.0.2
- name: datacube-ows
  url: https://github.com/opendatacube/datacube-ows
  version: 1.8.28
```

https://github.com/opendatacube/odc-stats/blob/5403a112465928d702546b761b8fc6d6a105dfea/odc/stats/io.py#L415

omad:
Fixed by #71

PR #71, opened 2022-11-03
use git commit sha as Patch in version

emmaai:
- auto update `Patch` in version `Major.Minor.Patch` to be git commit sha, e.g., version after this pr will be changed to `1.0.dev1ec0dce`
- use package version meta data rather than relying on `__version__` when generating processing metadata yaml. It can be very misleading and hard to figure exactly which version is used when `__version__` is not updated in repo.
- fix issue #65


odc/stats/io.py
```diff
@@ -408,21 +407,21 @@ def dump_with_eodatasets3(
...
 
         dataset_assembler.note_software_version(
-            "odc-stats", "https://github.com/opendatacube/odc-tools", __version__
+            "odc-stats",
+            "https://github.com/opendatacube/odc-stats",
+            version("odc_stats"),
         )
 
         dataset_assembler.note_software_version(
-            proc.NAME, "https://github.com/opendatacube/odc-tools", proc.VERSION
```
jmettes, line 419: Good catch. I thought this was already fixed at some point - must have been another one?

jmettes (approved):

SpacemanPaul (approved):

PR #73, opened 2022-11-14
Automate versioning, tagging and publishing of package

emmaai:
Changes:
- comply to [PEP440](https://peps.python.org/pep-0440/) using only numbers in the version of package
- automate the increment of `Patch` in version, tagging and publishing iff. there is any change in the code. The logic is fully tested with a minimal case here https://github.com/emmaai/test_github_actions
- combine publishing s3 and PyPI to share the same job of doing above
- solve issue #6 

It's supposed to achieve the match between package version and tag such that we can keep tack on the exact commit history.


.github/workflows/main.yml
```diff
@@ -40,10 +42,22 @@ jobs:
...
+        if: steps.verify-changed-version.outputs.any_changed == 'true'
         run: |
-          find . -name _version.py | xargs python ./scripts/patch_version.py \
-           $(git rev-parse --short HEAD)
+          # -1: auto increment of patch number
+          # other positive integers: set the patch number accordingly
+          python ./scripts/patch_version.py -1 ./odc/stats/_version.py
 
       - name: Build Clean Packages
         run: |
```
omad, line 77: Is there a missing step here that's supposed to push the wheels to a cache?

In the `test-wheels` job they're being pulled, but I can't see where from.

**Also**, there's a v3 of `actions/cache` that fixes a few deprecation warnings (and probably other things too).

omad (changes requested):
I need to think about this more. It seems like a couple of steps forward and maybe a step or two backwards.

I might simply be unfamiliar with thinking about doing continuous releases on every merge to `develop`...

Definitely check on the **cache** usage though please.

I don't like that it's bumping a version number to run a test, and then separately bumping the version number to save it into the git repository. I can see that it should work, but it seems confusing. Perhaps the two jobs should be merged, or, something...

.github/workflows/main.yml
emmaai: my understanding of `actions/cache`  -- it'll pull the cached file to `path` if it can find one and push from the `path` if it can't find one with `key`. So the lines here basically handle everything and it doesn't require explicit `push` or `pull` the cache.
https://github.com/opendatacube/odc-stats/blob/c62fabc497126331c4212e9c74fdf0517efe40fe/.github/workflows/main.yml#L26-L30

emmaai: As for the continuous release on every merge, based on the past behaviour how we update the version and build the docker image, most of time the version update and release happened when we wanted to change something in production, which basically implied every merge. We could potentially do it only when we think we need it, triggered by commit message syntax.

.github/workflows/main.yml
```diff
@@ -40,10 +45,33 @@ jobs:
...
+          if [[ $TAG == $CTAG || -z $CTAG ]]; then
+          # -1: auto increment of patch number
+          # other positive integers: set the patch number accordingly
+          python ./scripts/patch_version.py -1 odc/stats/_version.py
+          TAG=$(cat odc/stats/_version.py  | sed 's/[^0-9|.]//g')
+          git config --global user.email "$GITHUB_ACTOR@auto.noreply"
+          git config --global user.name "$GITHUB_ACTOR"
+          git commit -a -m "Bump version to $TAG"
+          git push origin HEAD:${{ github.head_ref }} --follow-tags
+          fi
```
emmaai: This segment is supposed to guarantee the version only got bumped up once no matter how many pushes or update on PR. It does multiple bumps because there isn't a tag yet. I'd be lazy on this, since the initial tag can be whatever as long as it matches `_version.py`.

.github/workflows/main.yml
```diff
@@ -40,10 +44,34 @@ jobs:
...
+        if: steps.verify-changed-version.outputs.any_changed == 'true'
+          && !(github.ref == 'refs/heads/develop')
+        id: patch-version
         run: |
-          find . -name _version.py | xargs python ./scripts/patch_version.py \
-           $(git rev-parse --short HEAD)
+          CTAG=$(git ls-remote --tags --sort='version:refname' | \
+            tail -1 | cut -d '/' -f3)
+          TAG=$(cat odc/stats/_version.py  | sed 's/[^0-9|.]//g')
+          if [[ $TAG == $CTAG || -z $CTAG ]]; then
```
Ariana-B, line 65: I feel like this logic could use some more comments

omad: I'm going to go further, and say that I don't think we want this version patching happening here.

I created a new branch based on this PR (#73), and was very surprised when I went to push to it a second time and found it had changed.
This Step had bumped the version number on me, inside a private branch, and will continue to bump the version number each time I push.

.github/workflows/main.yml
emmaai: @omad the comment here. re
> I created a new branch based on this PR (#73), and was very surprised when I went to push to it a second time and found it had changed.
This Step had bumped the version number on me, inside a private branch, and will continue to bump the version number each time I push.

.github/workflows/main.yml
emmaai: tag you on the comment I made above. yes, it's expected and explained in the comment.

.github/workflows/publish-new-version.yml
```diff
@@ -0,0 +1,106 @@
...
+          mkdir -p ./pips
+          ./scripts/mk-pip-tree.sh ./wheels/clean/ ./pips
+          find ./pips -type f
+
+      - name: Upload to S3
+        run: |
+          aws s3 ls "${S3_DST}"
+          aws s3 sync ./pips/ "${S3_DST}"
+        env:
+          AWS_ACCESS_KEY_ID: ${{ secrets.AWS_ACCESS_KEY_ID }}
```
jmettes, line 102: Can this be done via role? (e.g., like ECR pushing)

.github/workflows/publish-new-version.yml
```diff
@@ -0,0 +1,106 @@
...
+    steps:
+      - uses: actions/checkout@v3
+
+      - name: Push and tag versions
+        id: push-new-tag
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          GITHUB_ACTOR: autoaction
+        run: |
+          CTAG=$(git ls-remote --tags --sort='version:refname' | \
```
jmettes, line 26: As Ariana mentioned above, if you do add some comments, add them here too

tests/init_db.sh
```diff
@@ -4,8 +4,8 @@
 datacube system init
 datacube system check
 
-datacube product add https://raw.githubusercontent.com/digitalearthafrica/config/master/products/esa_s2_l2a.odc-product.yaml
-stac-to-dc --bbox='4,5,5,6' --collections='sentinel-s2-l2a-cogs' --datetime='2019-09-01/2020-01-01' --limit 1000
+# datacube product add https://raw.githubusercontent.com/digitalearthafrica/config/master/products/esa_s2_l2a.odc-product.yaml
```
jmettes, line 7: FYI I think Damien fixed these failing datasets in [odc-tools by mocking them](https://github.com/opendatacube/odc-tools/commit/2b57d1594574b46c15203492dff72cb88cbce6d1#diff-55e5c5a9e1d9db731fbe83c8f3ea8d424f4d592e1a661b6c5e5008f78964af0cR63)

.github/workflows/publish-new-version.yml
emmaai: can, once the terraform pipeline fixed I can add more access to `github-actions-role`. Not saying it wasn't but I haven't checked

tests/init_db.sh
emmaai: one important point of using the real datasets here is to guarantee the sanity/match of both end, code and input, and alarmed by any violation. To mock the dataset and pass the basic test was done enough (or should have) in the unit test.

omad (approved):
Okay, lets just merge this, and see how it goes.

What's the worst that can happen. 🙃 🔥🤣

PR #76, opened 2022-11-22
Fix linting errors

omad:
Merge AFTER #73


omad:
Okay, this is unexpected @emmaai 

Because the 'bump version' GHA run executes after the 'Test Code' GHA, and selects a different set of steps to run
image

image

**It completely masks the failing tests in the PR status!**

image
image

omad:
Also, another drawback of storing the version number in a text file, and incrementing it in every new branch, is that it'll almost always produce conflicts.
image

emmaai:
> Because the 'bump version' GHA run executes after the 'Test Code' GHA, and selects a different set of steps to run

that's a problem, ok, lemme figure out how to fix it.

emmaai:
The only way I could think of without conflict is to push the version update to the default branch, which can't be done with a protecting branch. As long as a commit is pushed to the remote branch not from current local, it'll require a `rebase` for further pushing. Any suggestions on a solution?

PR #77, opened 2022-11-28
Remove hardcoded version

emmaai:
changes:
- removed hardcoded `_version.py` and opted for `setuptools_scm` to version the package by GitHub tags
- removed build and test of wheels from pushing on `develop` branch, only does it with pr merging.
- added `MANIFEST.in` as required by `setuptools_scm` in building src dist, ref: https://pypi.org/project/setuptools-scm/


.github/workflows/main.yml
```diff
@@ -2,77 +2,55 @@ name: Run Code Checks
...
       - '!README.md'
+  workflow_run:
+    workflows: ["Create a new tag"]
+    types:
+      - completed
 
 jobs:
   build-wheels:
+    if: |
+      github.event_name == 'push' || (github.ref == 'refs/heads/develop'
```
Ariana-B, line 22: Would `github.ref == 'refs/heads/develop'` ever be true considering you have `branches-ignore: [develop]`?

Ariana-B (commented):
`branches-ignore: [develop]` really gives me pause (wouldn't that also encapsulate merges?) but otherwise lgtm

.github/workflows/main.yml
emmaai: `on` -- events operates in `or` logic, so the workflow here will be trigger either by `push` or `workflow_run`. If `push` happens on `develop`, it won't be triggered. Otherwise if `workflow_run` happens, it means a pr got merged and the commit got tagged, then the workflow should be triggered on `develop` since there is no restriction on branches for `workflow_run` such that it can build a "release" version based on the tag on `develop` branch.

emmaai: which implies, only pr merge will trigger a release but the commits caused by pr merge won't, that was I intended for.

Ariana-B: Makes sense! In that case it's all good by me

Ariana-B (approved):

PR #80, opened 2023-01-12
make frequency all literal when input temporal range

emmaai:
To fix an issue raised by @supermarkion that when `temporal_range=YY-MM--PNY` is given, the datasets are still binned by `key=YY--P(N+1)Y`. It will result in the wrong metadata. Hence the change is to accommodate both situations where it requires "all datasets available till today" or "the datasets within the given period ".

### DIFF
--- a/odc/stats/tasks.py
+++ b/odc/stats/tasks.py
@@ -324,7 +324,10 @@ class SaveTasks:
         msg(f"Total of {n_tiles:,d} spatial tiles")
 
         if self._frequency == "all":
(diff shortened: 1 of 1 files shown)


supermarkion (approved):
All good to my side. Thanks for your help, Emma.

fangfy:
hi @emmaai @supermarkion, I've just merged in another change to add different binning options. The problem you fixed here is also relevant, can we merge this PR?

PR #83, opened 2023-02-06
Fix xarray groupby

emmaai:
Changes

- remove `xarray` version constraint
- fix the tests
- patch the docker image

I don't have a sensible way to test the docker image against the current code since it's supposed to be a regression test and nothing should be broken. Maybe I shouldn't have updated the docker image in this pr. I'll leave the docker image test failed until next pr specially on it.

### DIFF
--- a/odc/stats/io.py
+++ b/odc/stats/io.py
@@ -229,8 +229,7 @@ class S3COGSink:
 
         ows_style = StandaloneStyle(ows_style_dict)
         # assign the time to xr.Dataset cause ows needs it
-        time_da = xr.Dataset({"time": time})
-        dst = ds.expand_dims(time=time_da.to_array())
+        dst = ds.expand_dims(dim={"time": [time]})
         img = dask.delayed(apply_ows_style)(ows_style, dst)
         return img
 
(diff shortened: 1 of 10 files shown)


woodcockr:
@emmaai I can’t review this PR (outside my knowledge of the code) but I did want to say thanks for getting on top of this groupby issue and I’m looking forward to the update!

odc/stats/io.py
```diff
@@ -229,8 +229,7 @@ def _apply_color_ramp(
 
         ows_style = StandaloneStyle(ows_style_dict)
         # assign the time to xr.Dataset cause ows needs it
```
Ariana-B, line 231: Seems like this comment is now outdated?

Ariana-B (approved):

odc/stats/io.py
emmaai: the code still does what's in the comment, only differently, to be compatible with newer xarray

SpacemanPaul (approved):
All looks sensible to me.

I think the docker requirements might more work given the remaining failing GH action - although I can't see what is going wrong in the docker build.

PR #85, opened 2023-02-10
upgrade packages version in the docker image

emmaai:
- remove the pin on `gdal` family, we'd know if any version causes trouble by test workflow. The consistency is guaranteed by enforcing the transform code here https://github.com/GeoscienceAustralia/dea-config/pull/1096
- put into bounds for `xarray` and `shapely`, not back compatible, `sqlalchemy<2.0` required by `datacube-core`, `gdal==3.6.2` has issue on reading from `s3`

### DIFF
--- a/docker/env.yaml
+++ b/docker/env.yaml
@@ -4,7 +4,7 @@ channels:
 dependencies:
   - python=3.8.10
   - libgdal
-  - gdal
+  - gdal<3.6.2
   - proj
   - rasterio>=1.3.2
   - libpq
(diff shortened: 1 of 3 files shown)


SpacemanPaul (approved):
We could maybe pin gdal `!=3.6.2` rather than `<3.6.2`?  Otherwise looks good.

emmaai:
> We could maybe pin gdal `!=3.6.2` rather than `<3.6.2`? Otherwise looks good.

not sure if they'd fix the problem in next version `>3.6.2` though. Just to be safe.

PR #90, opened 2023-03-16
Updated gm plugin to make it faster to run for S2

fangfy:
Changes made to remove hard-coded contiguity check and allow multiple cloud classes to be masked in the same data read.

This is still slower (~50% longer) and requires more memory to run than the version from before Jun 2022. Suggestions appreciated to improve the efficiency.

Tried to allow the case when cloud_filters is provided in the previous format but not sure if this will pass all tests.

### DIFF
--- a/odc/stats/plugins/gm.py
+++ b/odc/stats/plugins/gm.py
@@ -24,7 +24,9 @@ class StatsGM(StatsPluginInterface):
         mask_band: str,
         contiguity_band: Optional[str] = None,
         nodata_classes: Optional[Tuple[str, ...]] = None,
-        cloud_filters: Dict[Union[str, Tuple[str, ...]], Iterable[Tuple[str, int]]] = None,
+        cloud_filters: Dict[
+            Union[str, Tuple[str, ...]], Iterable[Tuple[str, int]]
+        ] = None,
         basis_band=None,
         aux_names: Dict[str, str] = None,
         resampling: str = "nearest",
--- a/odc/stats/plugins/gm.py
+++ b/odc/stats/plugins/gm.py
@@ -5,7 +5,7 @@ from typing import Optional, Union, Tuple, Iterable, Dict
 import xarray as xr
 from odc.algo import geomedian_with_mads
 from ._registry import StatsPluginInterface, register
-from odc.algo import enum_to_bool, keep_good_only
+from odc.algo import enum_to_bool, erase_bad
 from odc.algo import mask_cleanup
 import logging
 
@@ -104,7 +104,7 @@ class StatsGM(StatsPluginInterface):
             xx = xx.drop_vars([self._mask_band] + [self._contiguity_band])
         else:
             xx = xx.drop_vars([self._mask_band])
-        xx = keep_good_only(xx, ~bad)
+        xx = erase_bad(xx, bad)
         return xx
 
     def reduce(self, xx: xr.Dataset) -> xr.Dataset:
--- a/odc/stats/plugins/gm.py
+++ b/odc/stats/plugins/gm.py
@@ -88,10 +88,9 @@ class StatsGM(StatsPluginInterface):
 
         if self.cloud_filters is not None:
             for cloud_class, c_filter in self.cloud_filters.items():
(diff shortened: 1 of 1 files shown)


emmaai (approved):

emmaai:
The acceptable format of `cloud_filter` is here
https://github.com/GeoscienceAustralia/dea-config/blob/a0736288da43f8e1849a5ba09280e2c17a7068ab/dev/services/odc-stats/geomedian/ga_ls8c_nbart_gm_cyear_3.yaml#L3-L7

emmaai:
Had a look, these lines will drag slow the processing
https://github.com/opendatacube/odc-stats/blob/fe5f390013b0c034ec90d80d961a47c56492433c/odc/stats/plugins/gm.py#L85-L97

It's to accommodate the need to dilate cloud and shadow with different pixels. For LS, the expense is necessary.  For sentinel 2 I can see 4 of them got the same pixels. So I guess the change made in this pr was meant to achieve the previous behaviour. It might make dilation faster by reducing the graph earlier, but it wouldn't make memory usage less, neither would it speed up the loading.

Re: memory
In the code prior 06-2022, the mask band was loaded twice, once for masking, once for dilation. So the memory was released earlier than gathering and computing everything in one big graph (which is what happens with the current code).

Re: processing time
I got some guesses, several things could be changed see if it helps:
- drop the masking bands from `xx` first, which is not necessary by itself, but it'll enable the following changes
- `erase_bad` for every `bad` instead
- or `keep_good_only(..., invert=True)`, which is to invert by `numexpr` rather than `numpy`. I'm not sure how much faster, but the bottomline is that it should be.

odc/stats/plugins/gm.py
```diff
@@ -76,24 +78,31 @@ def native_transform(self, xx: xr.Dataset) -> xr.Dataset:
...
+            non_contiguent = xx.get(self._contiguity_band, 1) == 0
+            bad = bad | non_contiguent
+
         if self.cloud_filters is not None:
             for cloud_class, c_filter in self.cloud_filters.items():
-                cloud_mask = enum_to_bool(mask, (cloud_class,))
+                if isinstance(cloud_class, tuple):
+                    cloud_mask = enum_to_bool(mask, cloud_class)
+                else:
+                    cloud_mask = enum_to_bool(mask, (cloud_class,))
```
Kirill888: I usually prefer to "normalize" type of the variable using an if statement and then only have one "action line" afterwards when this is possible.

```suggestion
                if not isinstance(cloud_class, tuple):
                    cloud_class = (cloud_class,)
                cloud_mask = enum_to_bool(mask, cloud_class)
```

we are not doing something completely different when `cloud_class` is tuple vs string, we just treat a string and a tuple containing a single string as equivalent inputs.

odc/stats/plugins/gm.py
```diff
@@ -22,8 +22,9 @@ def __init__(
         self,
         bands: Tuple[str, ...],
         mask_band: str,
+        contiguity_band: Optional[str] = None,
         nodata_classes: Optional[Tuple[str, ...]] = None,
-        cloud_filters: Dict[str, Iterable[Tuple[str, int]]] = None,
+        cloud_filters: Dict[Tuple[str, ...], Iterable[Tuple[str, int]]] = None,
```
Kirill888:
```suggestion
        cloud_filters: Dict[Union[str, Tuple[str, ...]], Iterable[Tuple[str, int]]] = None,
```

from code below it looks like keys could be either a single string or a tuple of any number of strings.

fangfy:
Thanks @emmaai @Kirill888. To clarify, without this PR, the code is several times slower to run and uses even more momery. Main culprit for excessive memory usage is the unnecessary contiguity check. Doing the cloud masking in one iteration instead of many definitely speed up the code significantly. I understand the logic for wanting to do different dilation for different flags, hence happy to keep the option to provide multiple flag and filtering sets and wanted to make sure the landsat config still works.

I'll try `keep_good_only(..., invert=True).`

fangfy:
I don't think I can make the code to run faster at this point, so I'm merging the PR.

PR #93, opened 2023-04-18
Add the S2 GM to integration test case

supermarkion:
Add the DEA S2 GeoMED test case to odc-stats integration test. 

The DEA S2 GM can only work with the following cfg file: https://raw.githubusercontent.com/GeoscienceAustralia/dea-config/feature/add-S2ab-GM-processing-cfg/dev/services/odc-stats/geomedian/ga_s2ab_gm_4fyear_3.yaml. In this config file, it use DEA S2 ARD band names, use oa_s2cloudless_mask as Quality Control mask with 0 dilation.


emmaai (approved):
LGTM, as long as you uploaded your geotiff (to be compared to against) to s3

PR #96, opened 2023-06-30
Expand query for virtual product

emmaai:
- to reuse `_find_dss` in virtual product
- pin `datacube != 1.8.14` because of the issue in `1.8.14`

### DIFF
--- a/odc/stats/tasks.py
+++ b/odc/stats/tasks.py
@@ -177,11 +177,12 @@ class SaveTasks:
         query: Dict[str, Any],
         dataset_filter=None,
         predicate=None,
+        fuse_dss: bool = True,
     ):
         """
         query and filter the datasets with a string composed by products name
         A string joined by `-` implies union of all datasets
-        A string joined by `+` implies intersect (filtered and then groupby) time
+        A string joined by `+` implies intersect (filtered and then groupby) against time
         return a generator of datasets
         """
         if dataset_filter is None:
@@ -201,9 +202,9 @@ class SaveTasks:
             3. Multiple different separators between the product names is respected by left-right order
             e.g., ga_ls8+-ga_ls7 -> separator is `+` as `+` proceeds `-` from left to right
             """
-            product_list = re.split(r"\+|-", products_str)
+            product_list = re.split(r"[\+-]{1,}", products_str)
             product_list = list(filter(None, product_list))
-            group_size = len(re.findall(r"\w\+{1,}[\w-]", products_str))
+            group_size = len(re.findall(r"\w{1,}\+[-\+]{0,}\w{1,}", products_str))
             return product_list, group_size
 
         product_list, group_size = sanitize_products_str(products)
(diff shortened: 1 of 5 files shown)


odc/stats/tasks.py
```diff
@@ -151,18 +158,65 @@ def match_dss(groups, group_size):
...
+            dataset_filter = {}
+
+        # query by a list of products is not a "officially" supported feature
+        # but it is embedded in the code everywhere
+        # mark it for ref
+
+        product_list = re.split(r"\+|-", products)
+        product_list = list(filter(None, product_list))
+        group_size = len(re.findall(r"\+", products))
+
```
SpacemanPaul, line 211: I don't understand what's going on here.  What is the expected format of "products"?  That's some weird stuff going on with pluses and minuses.  

`products="prod1+prod2-prod3+prod4"` and `products="+++prod1+prod2-+-prod3++prod4"` both  yield the same 4 member `product_list`, but one givess a `group_size` of 4 and the other 7??  With no warning/error message?

It seems like the main purpose of `find_dss` is to expand a product string into a product list, so a little bit of documentation here would be helpful.

(It's still better than it was, mind you.)

odc/stats/tasks.py
```diff
@@ -151,18 +158,65 @@ def match_dss(groups, group_size):
...
+            def map_fuse_func(x):
+                return fuse_ds(*x, product=fused_product)
+
+            dss = map(map_fuse_func, grouped_dss)
+            return dss
+        else:
+            return grouped_dss
+
+    @classmethod
+    def _find_dss(
```
SpacemanPaul, line 173: Having `_find_dss` and `_get_dss` side by side like this is potentially confusing.

setup.cfg
```diff
@@ -22,7 +22,7 @@ install_requires =
     botocore
     click>=8.0.0
     dask
-    datacube
+    datacube<=1.8.13
```
SpacemanPaul: As this issue has been fixed in develop and will therefore definitely be fixed in 1.8.15, maybe just require `!=1.8.14`

SpacemanPaul (changes requested):
Good refactor, but a little more documentation (and/or validation) would be helpful.

Also this patch has only has like 15% coverage. A quick coverage run of  `find_dss` with mocks should be fairly quick to throw together.

odc/stats/tasks.py
```diff
@@ -151,18 +158,85 @@ def match_dss(groups, group_size):
...
+        def sanitize_products_str(products_str):
+            """
+            split a string composed by product names into a list of product names
+            e.g., ga_ls8-ga_ls7 -> [ga_ls8, ga_ls7]
+            rules:
+            1. Any separator (`+/-`) at the start or the end is disregarded
+            2. Multiple same separators between the product names are treated as one
+            3. Multiple different separators between the product names is respected by left-right order
+            e.g., ga_ls8+-ga_ls7 -> separator is `+` as `+` proceeds `-` from left to right
+            """
```
SpacemanPaul, line 204: This is much better but I still don't understand the difference between + and - as a separator or what the significance of `group_size` is.

SpacemanPaul (approved):
The product string splitter could do with a just a bit more documentation (specifically around the "grouping" concept).  Otherwise looking great. Thanks Emma.

odc/stats/tasks.py
emmaai: I'll throw in a small segment in readme in next pr. It's meant to support the concept of `juxtapose` and `collate` as well as the similar idea in `odc-stats` with the same set of code.

Issue #100, opened 2023-10-13
misplaced chunk in  geomedian with dask


emmaai:
"Misplaced" chunks in geomedian as shown in the picture below. My current theory is that the graph unpack is triggered by `persist` before the graph built is completed, then something goes wrong in the middle. It's either a bug in `dask`, or we shouldn't do this at all. It seems to happen randomly. I don't have a reliable way to reproduce it.
![misplaced_chunk](https://github.com/opendatacube/odc-stats/assets/5643062/7aa30ea5-a650-4c3e-86a2-736f32b5634f)

robbibt:
Hmm, a user recently emailed us about a kind of similar looking issue:

![image](https://github.com/opendatacube/odc-stats/assets/17680388/eb0a17e1-929c-44c4-b683-f8c447e2781e)

They're not doing geomedians, but are doing large-scale medians/means/stdev using Dask. The previous assumption was that this was caused by the S3 random data access problem we've talked about previously on Teams, but I wonder if it might actually be related to this issue too....

(or maybe it's completely different, hard to tell from those screenshots)

emmaai:
can confirm none of above fixed the issue, it happened again in 2015 geomedian test processing. http://dea-public-data-dev.s3-website-ap-southeast-2.amazonaws.com/?prefix=test/gm-ls8-dilation-6-cloud-opening-5-v2/3-0-0/x40/y13/2015--P1Y/

emmaai:
fixed by https://github.com/opendatacube/odc-algo/pull/4

SpacemanPaul:
Thanks for your work on this @emmaai ! I know this was a tough one.

PR #101, opened 2023-10-25
remove early dask unpack

emmaai:
It's to test/address one of my theories on issue #100:  the early unpack triggered something unexpected in dask graph.

My second theory is that the node id in dask graph is not unique when there are many chunks.


omad (approved):
Sure, give it a go.

Why do you need to commit the change to test it out though?

emmaai:
I’ve tested but I’m not sure if it solved the issue or it didn’t hit the
odd. So I wanna test it in a significant large scale in the coming summary
products 2015 reprocessing test.

PR #114, opened 2024-01-16
Handle credentials rotation

emmaai:
Changes:

- removed the static credentials, let `boto3.client` handle the credentials rotation
- changed how `s3_client` (ref:https://github.com/opendatacube/odc-tools/pull/597) is managed on dask cluster, one for scheduler and one per work (among all threads rather than each per thread)


SpacemanPaul (approved):
Looks good, but obviously can't be merged until the changes in odc-cloud are released.

PR #115, opened 2024-01-17
update odc-cloud version and install new odc-stats

emmaai:
as the title ref: pr #114


SpacemanPaul (approved):
Sorry for the delay getting to this PR.  I had a crazy day yesterday.

PR #124, opened 2024-04-15
Construct tasks with stac

emmaai:
As the tile.

- take s3 paths and product names as input to construct task
- enable mixed inputs of s3 paths and product names
- fuse datasets and products properly when the datasets are constructed from stac

relevant change:
https://github.com/opendatacube/odc-stac/pull/152

### DIFF
--- a/odc/stats/tasks.py
+++ b/odc/stats/tasks.py
@@ -123,8 +123,6 @@ def sanitize_products_str(products_str):
     2. Multiple same separators between the product names are treated as one
     3. Multiple different separators between the product names is respected by left-right order
     e.g., ga_ls8+-ga_ls7 -> separator is `+` as `+` proceeds `-` from left to right
-    4. It can NOT deal with the different combo of "+/-"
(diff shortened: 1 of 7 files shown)


odc/stats/tasks.py
```diff
@@ -187,53 +229,92 @@ def _find_dss(
...
-            **query,
-        )
+        indexed_products = []
+        non_indexed_products = []
+        for p, indexed in product_list:
+            if indexed:
+                indexed_products += [p]
+            else:
+                non_indexed_products += [p]
+
```
SpacemanPaul, line 249: I would use `arr.append(p)` here rather than `arr += [p]`.   Probably compiles to the same byte code, but I think it's easier to read.

SpacemanPaul (approved):
Looks good to me.

odc/stats/tasks.py
emmaai: It's only my preference to type less. :) also, `+=` is more friendly if appending all the items in another list. I'd agree with `append` easier to read though.

PR #126, opened 2024-04-19
adapt to package name convention change

emmaai:
as the title, e.g., src package name changed from `odc-stats.version.tar.gz` to `odc_stats.version.tar.gz`


omad:
Why the change?

omad (approved):

emmaai:
> Why the change?

Idk, `python setup.py sdist` suddenly decided to follow the naming convention of `pythons setup.py bdist_wheel`

PR #129, opened 2024-05-29
Landcover plugin the maximum consecutive months of a pixel being vegetation

emmaai:
The logic is described as follows:

- FC bands are masked by `valid:=0x000x_xx00` & `ue < 30` where `x` means any value of `{0, 1}`
- WO band are classifies as `wet:=0x1000_0x00`, hence the valid wet+dry pixels are `wet | valid`
- FC bands grouped by solar day with mean, WO bands on `nodata=np.nan`
- Monthly medians are taken over FC bands and `wet` band for further classification of veg vs. non-veg
- `veg:= pv>bs | npv > bs & wet <= 0`, where `wet > 0` implies that at least half of observations  in a month pixels being `wet`, otherwise non-veg. `nan := nodata`.
- count and save the maximum consecutive months that pixels being veg, where `nan` (or `nodata`) is considered to be neither veg nor non-veg. Though it doesn't break consecutiveness either, e.g., a sequence of `{1, nan, 1, 1}` is deemed as `3` consecutive months, where `1` indicates veg, while a sequence of `{1, 0, 1, 1}` is deemed as `2` consecutive months as `0` breaks consecutiveness and hence number is reset.

Note:

pylint hook in pre-commit ci is out of date and it failed in python 3.12. I don't see any way to fix it, hence made pylint a local hook and skip in ci.


odc/stats/plugins/lc_fc_wo_a0.py
```diff
@@ -0,0 +1,206 @@
+"""
+Plugin of Module A0 in LandCover PipeLine
```
JM-GA, line 2: Hi @emmaai, is it worth renaming this file to be called something based on the output bands - thinking maybe users will not know what module a0 refers to. Alternatively maybe adding the architecture diagram to the PR and a reference to the output of module

SpacemanPaul: Maybe  `lc_fc_wo_stats.py` ?

SpacemanPaul (approved):
Agree with @JM-GA that some of the naming could be improved, particularly the a0.  But happy to just flag for revisit given further work on this module is anticipated.

emmaai:
I have marked the awful name as an issue to be resolved along development progressing

tebadi (commented):
I had a look after it was merged but it looks good!

PR #132, opened 2024-06-05
Fix mangroves and wo plugins

emmaai:
- enforce the use of mangroves extent shapefile
- set `nodata=np.nan` explicitly for wo float band as esri doesn't agree with the default nan on float type.

Note:
nodata issue of gm to be fixed in `odc-algo`. The virtue here is that every band (variable) should carry `nodata` as one of attributes no matter what data type it is, not claiming it's a "better" solution.


SpacemanPaul (approved):
Looks straightforward and sensible to me.

tebadi (approved):

Issue #131, opened 2024-06-05
Read product definition for bands instead of hard coding


emmaai:
Minimum:
- verify the variable names against bands in product definition to ensure the correct output band/file names.
- read the data type and `nodata` from product definition.

PR #135, opened 2024-06-13
Fix gm nodata

emmaai:
I changed my mind on where to fix gm `nodata`  issue, based on the following reasons:
- Both s2 and ls inherit the same plugin and hence same function in computing the gm. They can have their different ways in defining `nodata` for extra bands other than the input bands. 
- The `nodata` is defined  in production definition yaml. All datasets should respect and follow the same definition such that `dc.load` and `rioxarray.open` will result the same `nodata` information. For future improvement ref issue #131 
- The same "hardcode" approach also caused discrepancy between data type of GeoTIFF and one in product def, e.g., `count` being `uint16` in GeoTIFF while `int16` in product def, which results in `int16` in `dc.load` while `uint16` with other tools
- gdal doesn't fully respect default `nodata=nan` for float when generating overviews either. Thus it'd be good that we honour `nodata` definition every level. In this PR, I allowed `nodata` not defined for `count` band because of a) data type discrepancy mentioned above, b) `nodata` not presents in raster hence no trouble atm

I decide to address the coding issues later while keep the path clear for now.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -33,7 +33,7 @@ jobs:
         with:
           fetch-depth: 0
 
-      - uses: actions/cache@v3
+      - uses: actions/cache@v4
         id: wheels_cache
         with:
           path: ./wheels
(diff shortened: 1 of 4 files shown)


JM-GA (approved):

SpacemanPaul (approved):
LGTM

PR #138, opened 2024-06-26
extension for landcover pipeline and plugins

emmaai:
Major changes:
- generalize datasets fusion over `>=2` products
- allow time dimension to be ignored for certain products and bands, e.g., masking yearly dataset with a multi-year product
- allow products and bands to be optional when aligning/matching over datasets
- new plugin for landcover l3/l4 staging results, the logic of algorithm is shown in the image below
image

Minor changes:
- change the default value of click to be none, for better hierarchy of options overwriting, i.e, command line option > yaml option > default option


odc/stats/utils.py
```diff
@@ -299,7 +299,7 @@ def dedup_s2_datasets(dss):
     return out, skipped
 
 
-def fuse_products(type_1: DatasetType, type_2: DatasetType) -> DatasetType:
+def fuse_products(*ds_types) -> DatasetType:
```
SpacemanPaul, line 302: `datacube.model.Product` has been available in core as a more readable/less confusing synonym of `DatasetType` for quite some time.

SpacemanPaul (approved):

odc/stats/utils.py
emmaai: lemme mark it for next PR

JM-GA (approved):

PR #142, opened 2024-07-04
check datetime instead of assuming time and freq

emmaai:
As the title. It was coded based on the assumption that we'd only run year by year as a temporary solution before indexing. It's not true anymore, and it could well be multi-year summaries too.


tebadi (approved):

SpacemanPaul (approved):
Looks reasonable, but test failures will need to be addressed.

PR #143, opened 2024-07-18
plugin of urban model for landcover

emmaai:
Major change:
- add plugin for urban model
- make TensorFlow multithreading safely with dask worker and threads

The diagram is as below:
image
Minor change:
- update dependencies in docker image

Note: 
The performance/accuracy of model itself is not guaranteed in this implementation. It was and will be dealt with separately.


SpacemanPaul (approved):
LGTM

PR #146, opened 2024-08-06
Woody cover and cultivated model plugins for landcover

emmaai:
Major changes:
- added dask worker plugin for `treelite` predictor for multithreading safety
- added a plugin class for all models of `treelite` as they share common routines
- added plugins for woody cover and cultivated model
- added tests for above changes

cultivated model diagram:
image

woody cover diagram
image

Minor changes:
- updated `odc-algo` head which concerns safe casting (ref: https://github.com/opendatacube/odc-algo/pull/13) and memory sanity (ref: https://github.com/opendatacube/odc-algo/pull/12)
- `docker-compose` to `docker compose`

### DIFF
--- a/.github/workflows/statistician-image.yml
+++ b/.github/workflows/statistician-image.yml
@@ -44,7 +44,7 @@ jobs:
         shell: bash
         run: |
           cd docker
-          docker-compose build
+          docker compose build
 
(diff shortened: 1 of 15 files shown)


SpacemanPaul (approved):
I don't really understand what's happening here tbh, but looks clean enough.

PR #149, opened 2024-08-23
Fix bugs in cultivated and woody cover plugins

emmaai:
Addressed issues:
- results from cultivated and woody cover run with pipeline showed discrepancy on tiles randomly from run to run, after investigations and experiments, confirmed that it's caused by `numexpr` package, the relevant issue was raised https://github.com/pydata/numexpr/issues/494
- input bands missing due to unavailable inputs from even upper streams, hence allow the graceful exist.

Note:
I have put into a hot (means very hacky) fix for the `numexpr` package, which is remarked in the docker file. We must follow up on this issue and address in the future.


SpacemanPaul (approved):
All looks reasonable to me.  Hopefully we get prompt action from the folks at numexpr.

PR #151, opened 2024-09-12
Added a plugin for processing level3 Land Cover product.

tebadi:


odc/stats/plugins/lc_level3.py
```diff
@@ -0,0 +1,54 @@
...
+    def measurements(self) -> Tuple[str, ...]:
+        _measurements = ["level3_class"]
+        return _measurements
+
+    def reduce(self, xx: xr.Dataset) -> xr.Dataset:
+
+        l34_dss = xx.classes_l3_l4
+        urban_dss = xx.urban_classes
+        cultivated_dss = xx.cultivated_class
+
```
emmaai, line 28: could you either give the integer a name or comment what the number means?

emmaai (changes requested):
Could you please enable precommit in your local repo and make sure all the lint pass before committing?

odc/stats/plugins/lc_level3.py
```diff
@@ -0,0 +1,54 @@
...
+        cultivated_mask = l34_dss == 110
+        l34_cultivated_masked = xr.where(cultivated_mask, cultivated_dss, l34_dss)
```
emmaai, line 32: cultivated is masked by the veg in input already, so it's just matter of merging valid data (`data != nodata`). Thought the logic pretty much the same, it's better to reflect that point and avoid the confusion.

emmaai (approved):
Thanks!

PR #155, opened 2024-09-26
Add water frequency to A1 module.

tebadi:


odc/stats/plugins/lc_veg_class_a1.py
```diff
@@ -205,17 +209,41 @@ def l3_class(self, xx: xr.Dataset):
...
-        return l3_mask
+        # Now add the water frequency
+        # Divide water frequency into following classes:
+        # 0 --> 0
+        # (0,0.25] --> 0.25
+        # (0.25,1] --> 1
+        water_seasonality = expr_eval(
+            "where((a > 0) & (a <= wt), wt, a)",
+            {"a": xx["frequency"].data},
+            name="mark_veg",
```
emmaai: name is not a big deal for dask task, but might mislead if there is an error.

odc/stats/plugins/lc_veg_class_a1.py
```diff
@@ -205,17 +209,41 @@ def l3_class(self, xx: xr.Dataset):
...
+            dtype="float32",
+            **{"wt": self.water_seasonality_threshold, "nodata": NODATA},
```
emmaai: `nodata` is not used in expression.

odc/stats/plugins/lc_veg_class_a1.py
```diff
@@ -205,17 +209,41 @@ def l3_class(self, xx: xr.Dataset):
...
+        )
+
+        water_seasonality = expr_eval(
+            "where((a > wt) & (a <= 1), 1, a)",
+            {"a": water_seasonality},
+            name="mark_water_seasonality",
+            dtype="float32",
```
emmaai, line 231: the output type should be `uint8` if it's final result, but  `nan` is not masked with `nodata=255`, so there should be an extra step to mark `nan` as `255` then output `uint8`

emmaai (changes requested):
data type and nodata requires some care.

emmaai (approved):
thanks!

JM-GA (approved):

PR #157, opened 2024-10-04
Land Cover A3 Module

tebadi:
The classes in this module were obtained from the following two sources and will match C2:

https://knowledge.dea.ga.gov.au/data/product/dea-land-cover-landsat/?tab=details

https://bitbucket.org/au-eoed/livingearth_lccs/src/1e78111ead2feb1179bf9fb30f6e8744aed65886/colour_schemes/lccs_l4_colour_scheme_golden_dark_au.csv


JM-GA (approved):
@tebadi Just a FYI - once the naming conventions are approved - we will likely need to rename some of the odc-stats modules - but this will be at a later date before we start bulk processing

emmaai (commented):
could you gather all this into `plugins.lc_cevel3`? I don't see a point of having another intermediate product. The final results can be produced in `plugins.lc_level3` already. Also the sub-classes of level 4 are mutually exclusive, there is no need to have different bands anymore.

emmaai (changes requested):
as we agreed on the redundancy.

PR #161, opened 2024-10-23
Add Level-3 and Level-4 bands for Land Cover product.

tebadi:
This change will create the level4 and level3 land cover bands.

The fix_l3 will be applied on top of this PR which will fix the nodata issue. This fix will be applied to Level4 as well.

Links to definitions and thresholds used to generate level-4 bands:

- https://bitbucket.org/au-eoed/livingearth_lccs/src/1e78111ead2feb1179bf9fb30f6e8744aed65886/colour_schemes/lccs_l4_colour_scheme_golden_dark_au.csv

- https://knowledge.dea.ga.gov.au/data/product/dea-land-cover-landsat/?tab=details


JM-GA (approved):

odc/stats/plugins/l34_utils/lc_water_seasonality.py
```diff
@@ -0,0 +1,40 @@
...
+        **{"watseas_trh": water_seasonality_threshold},
+    )
+    water_season_mask = expr_eval(
+        "where((a==watersea_nodata), 255, a)",
+        {"a": water_season_mask},
+        name="mark_water_season",
+        dtype="uint8",
+        **{
+            "watseas_trh": water_seasonality_threshold,
+            "watersea_nodata": WATER_FREQ_NODATA,
```
emmaai: the correct `nodata=255` from the raster, also it won't work as `water_season_mask` has been converted to `uint8` from last step, `-999` was lost had it been the effective `nodata`.

odc/stats/plugins/l34_utils/lc_level3.py
```diff
@@ -0,0 +1,50 @@
...
+        {
+            "a": res,
+            "b": xx.urban_classes.data,
+        },
+        name="mark_urban",
+        dtype="uint8",
+        **{"_u": 210},
+    )
+
+    intertidal_mask = expr_eval(
```
emmaai: this mask is redundant.

odc/stats/plugins/l34_utils/l4_bare_gradation.py
```diff
@@ -0,0 +1,60 @@
...
+from . import utils
+
+
+NODATA = 255
+
+
+def bare_gradation(xx: xr.Dataset, bare_threshold, veg_cover):
+
+    # Now add the bare gradation
+    fcp_nodaata = -999
```
emmaai: is it a typo? I don't see anywhere changing the fcp `nodata` value.

odc/stats/plugins/l34_utils/l4_bare_gradation.py
```diff
@@ -0,0 +1,60 @@
...
+    bs_mask = expr_eval(
+        "where(a<m, 100, a)",
+        {"a": bs_mask},
+        name="mark_veg",
+        dtype="uint8",
+        **{"m": bare_threshold[0]},
+    )
+    # Apply bare gradation expected output classes
+    # Map bare gradation classes
+    bs_mapping = {100: 10, 120: 12, 150: 15}
```
emmaai: it might be easier to allocate a new array instead of the extra mapping, also reduce the confusion. It's bit hard to follow that it's essentially to avoid the mixing of small and big values in the original array.

odc/stats/plugins/l34_utils/l4_veg_cover.py
```diff
@@ -0,0 +1,88 @@
...
+        "where((a>=m)&(a<n), 100, a)",
+        {"a": veg_mask},
+        name="mark_veg",
+        dtype="uint8",
+        **{"m": veg_threshold[4], "n": veg_threshold[5]},
+    )
+
+    # Define mapping from current output to expected a3 output
+    # Map vegetation cover classes
+    veg_mapping = {160: 16, 150: 15, 130: 13, 120: 12, 100: 10}
```
emmaai: same as `bare_gradition`

odc/stats/plugins/l34_utils/l4_water_persistence.py
```diff
@@ -0,0 +1,69 @@
...
+        "where((a>=m)&(a<n), 90, a)",
+        {"a": water_mask},
+        name="mark_water",
+        dtype="uint8",
+        **{"m": watper_threshold[0], "n": watper_threshold[1]},
+    )
+
+    # Apply water persistence expcted classes
+    # Map values to the classes expected in water persistence in land cover Level-4 output
+    waterper_wat_mapping = {100: 1, 70: 7, 80: 8, 90: 9}
```
emmaai: same as `bare_gradition`

emmaai (changes requested):
as in individual comment

odc/stats/plugins/l34_utils/lc_water_seasonality.py
tebadi: Good point, refactored the code to effectively replace "nodata"

odc/stats/plugins/l34_utils/lc_level3.py
tebadi: This mask is used in level4 water classification.

odc/stats/plugins/l34_utils/l4_water_persistence.py
tebadi: Deleted the mapping.

odc/stats/plugins/l34_utils/l4_veg_cover.py
tebadi: Deleted the mapping.

odc/stats/plugins/l34_utils/l4_bare_gradation.py
tebadi: Agree. I've modified the code and deleted the mapping here and  everywhere else.

odc/stats/plugins/l34_utils/l4_bare_gradation.py
tebadi: I've removed the redundant statement.

odc/stats/plugins/l34_utils/lc_level3.py
emmaai: I mean it's redundant in a sense that l3 and l4 doesn't have to be generated in that order. The water persistence and consequently l4 water classification can happen well ahead of l3 aggregation.  https://github.com/opendatacube/odc-stats/blob/40ab39861f9cec261360568bdc01807dc8ba0c7c/odc/stats/plugins/lc_level34.py#L92 and https://github.com/opendatacube/odc-stats/blob/40ab39861f9cec261360568bdc01807dc8ba0c7c/odc/stats/plugins/lc_level34.py#L107 doesn't depend on the aggregated results of l3, so l3 aggregation needs not to happen first.

tebadi: This is addressed now.

emmaai (approved):
let's roll see how it goes :)

PR #163, opened 2024-10-25
Fix the issue of input data nodata changing by the odc loading

emmaai:
As the title, ref https://github.com/opendatacube/datacube-core/issues/1646

- carry the nodata following the same convention in `datacube`
- minor refactor of code always reading `nodata` from dataset attributes

Note that it's not a fundamental solution, as the issue raised in `datacube`, ideally the loading process should inform any change, as `None` can be understood and dealt with so many ways. Currently `datacube` rely on the assumption that people agree with one, which might not always be true.

### DIFF
--- a/odc/stats/plugins/_base.py
+++ b/odc/stats/plugins/_base.py
@@ -53,7 +53,7 @@ class StatsPluginInterface(ABC):
                 xx[var].attrs.get("nodata") is None
                 and np.dtype(xx[var].dtype).kind == "f"
             ):
-                xx[var].attrs["nodata"] = np.nan
+                xx[var].attrs["nodata"] = xx[var].dtype.type(None)
         return xx
 
     def fuser(self, xx: xr.Dataset) -> xr.Dataset:
--- a/odc/stats/plugins/_base.py
+++ b/odc/stats/plugins/_base.py
@@ -2,6 +2,7 @@ from abc import ABC, abstractmethod
 from typing import Mapping, Optional, Sequence, Tuple
 
 import xarray as xr
+import numpy as np
 from datacube.model import Dataset
 from datacube.utils.geometry import GeoBox
 from odc.algo import to_rgba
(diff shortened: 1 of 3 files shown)


JM-GA (approved):

SpacemanPaul:
It looks like it should work, but I'm confused why we have to handle so many weird corner cases here (float, nodata=NaN; float, nodata=None; int, nodata=255; int, nodata=255 OR MORE, etc). Surely we have more control over the dataflow here?

emmaai:
> It looks like it should work, but I'm confused why we have to handle so many weird corner cases here (float, nodata=NaN; float, nodata=None; int, nodata=255; int, nodata=255 OR MORE, etc). Surely we have more control over the dataflow here?

The problem is with `product definition`. If the `Dataset` used in loading is created with odc database query, it has the correct `dtype` and `nodata`, that'd be `uint8` and `255`. If it is created by `stac`, then the `product definition` is missing and hence no `dtype` or `nodata`, in this case, odc falls to `float` and `None`. In the loading, `None` will be converted to `nan`, and `odc` loading will substitute the `nodata` in geotiff `metadata` with `nan`. 

Currently there is no ideal way to gain the information of product definition for datasets not indexed, dataset metadata and product definition seems totally irrelevant unless that we are always assumed to know where to find that piece of information.

All the corner cases are caused by how odc deals with missing information by default, as stated in the issue raised on `datacube`.  Hence the dataflow actually loses the control over the data loading, not knowing either product definition or the convention of `datacube` .

The essence of a solution would be linking product definition to dataset metadata such that the dataflow can know for sure what to be loaded without relying on `datacube` default. Though it requires some upgrades on dataset metadata, and further discussion on how to implement it.

tebadi (commented):
This is fixing nodata at level3 for the cultivated band which is a specific known case.  Level4 loads other bands and If nan is present in those bands, this won't fix that. Example are woody_cover, pv_pc_50 and water_frequency.

emmaai:
The change is not meant to fix any particular band or dataset. It only
informs what’s the real nodata value of the data. The individual plugin
should deal with the information accordingly, level 3 plugin only serves as
an example of how.

tebadi (approved):

PR #165, opened 2024-11-06
Change cultivated aggregation logic and make sensor choice an option

emmaai:
Changes:
- the logic of cultivated changed from `any sensor says cultivated -> cultivated` to `any sensor says natural -> natural` as bias towards cultivated class was observed in validation
- sensor choice as an option for both cultivated and woody cover
- woody cover band name aligned with official definition `woody` instead of `woody_cover`

### DIFF
--- a/odc/stats/plugins/lc_ml_treelite.py
+++ b/odc/stats/plugins/lc_ml_treelite.py
@@ -110,7 +110,7 @@ class StatsMLTree(StatsPluginInterface):
                     (self.chunks["x"], self.chunks["y"], -1, -1),
                     dtype="float32",
                     name=ds.type.name + "_yxbt",
-                )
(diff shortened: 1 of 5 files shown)


JM-GA (approved):

tebadi (commented):
Has the impact on the accuracy of validation data been measured? Not sure we have data from different sensors in the validation but if we do, it's worth checking this.

emmaai:
> Has the impact on the accuracy of validation data been measured? Not sure we have data from different sensors in the validation but if we do, it's worth checking this.

I have to merge this to be able to run another continental scale and then on validation data. There would be retrace potentially.

tebadi (approved):

tebadi (approved):

PR #166, opened 2024-11-12
Make output band name an option

emmaai:
As the tile, plugin can read `measurements` from config yaml in addition to implementing parent class property.


JM-GA (approved):

tebadi (approved):
Look good!

PR #176, opened 2024-12-05
correct veg/non logic in landcover veg frequency

emmaai:
As the title. It removed the "likely water" -> non-veg, introduced consecutive observations of non-veg < threshold(2) -> veg. The earlier caused bare soil along water edges, where wo summaries don't have clear enough data to tell for sure if it's water, consequently it'll be non-water, non-water + non-veg -> bare soil. The absence of later would cause similar problem, where a water - veg - water - veg sequence would cause a pixel to be bare soil. It's very rare to have soil - veg - soil - veg sequence after the sanity check on the results of first implementation, and intuitively. The condition was also used in c2, though back to the time,  there didn't seem to make any statistical difference to classify dry pixels either way. Now the pro of having the condition clearly out-weighted con, decide to introduce it now.


JM-GA (approved):

odc/stats/plugins/lc_fc_wo_a0.py
```diff
@@ -262,8 +236,30 @@ def reduce(self, xx: xr.Dataset) -> xr.Dataset:
...
+            dtype="uint8",
+            **{"nodata": NODATA},
+        )
+        max_count_non_veg = self._max_consecutive_months(non_veg_data, NODATA)
+
+        # non-veg < threshold implies veg >= threshold
+        # implies any "wet" area potentially veg
+
+        max_count_veg = expr_eval(
+            "where((a<_v)&(b<_v), _v, b)",
```
tebadi, line 257: Is this < or <=?

odc/stats/plugins/lc_fc_wo_a0.py
```diff
@@ -262,8 +236,30 @@ def reduce(self, xx: xr.Dataset) -> xr.Dataset:
...
 
-        data = self._veg_or_not(xx)
-        max_count_veg = self._max_consecutive_months(data, NODATA)
+        # consecutive observation of veg
+        veg_data = self._veg_or_not(xx)
+        max_count_veg = self._max_consecutive_months(veg_data, NODATA)
+
+        # consecutive observation of non-veg
+        non_veg_data = expr_eval(
+            "where(a<nodata, 1-a, nodata)",
```
tebadi, line 245: I don't understand why it's 1-a

emmaai: 1: veg 0: non-veg. after flipping 1: non-veg, 0: veg, easier to use same function to calculate monthly median.

odc/stats/plugins/lc_fc_wo_a0.py
emmaai: strictly less, it's specific for the pattern of `non-veg - veg - non-veg - veg`

tebadi: Thanks for confirming this

tebadi (approved):

PR #186, opened 2025-02-28
Major upgrade with `debian=12` and `python=3.12`

emmaai:
WIP


pjonsson:
Most other odc repositories seem to use Ubuntu, what caused the switch to Debian for this repository?

emmaai:
the trigger point is a) ubuntu 24.04 bloated too much, and b) mamba image doesn't differentiate with/out cuda since 24.04, which bloated the image even more.
fundamental reasons a) the docker image is only for production, not for extension of user case, unlike other packages/images, given this presumption b) Debian is leaner, faster, safer and compatible enough for production purposes.

pjonsson:
> b) Debian is leaner, faster, safer and compatible enough for production purposes.

If you haven't already done it, you might want to run Trivy or some other security scanner on the corresponding Ubuntu/Debian images for this PR and compare the results. A couple of years ago (sometime during Q2-Q3 in 2023), I did an apples to apples comparison for some other projects between Ubuntu LTS-based images and Debian-based images, and the Ubuntu LTS-based images had fewer vulnerabilities because Canonical were more inclined to provide security-updates for the packages. I haven't had time to make any comparison after that, so things might have changed in the last couple of  years.

emmaai:
It’s quite the opposite from what I got from Ubuntu 24.04 vs Debian 12.10.
Also it’s a wip, as long as it passes scan in aws ecr, it’s good enough for
production. Imo, those vulnerabilities exist in theory, but in the user
case of pipeline, the risk of exploitation is close to naught.

PR #189, opened 2025-03-26
Add support for "monthly" statistics

dtpc:
Add `"monthly"` to the list of valid inputs for the `--frequency` option and support binning by calendar month.


emmaai (approved):
thanks for the cleanup!

PR #195, opened 2025-06-13
upgrade to odc>1.9

emmaai:
As the tile. Current `develop` is tagged as `develop_1.8` for record.
- integrated changes made by @Ariana-B on the branch `integrate_1.9` wrt update on `odc.*>1.9`
- shuffled `numexpr` related functionality back to `odc-algo`
- adopted `native_loading` from `odc-algo` atm, further discussion on shuffling the code again

Note that:
- tests on code `run code check` should pass
- tests on the older docker image `statistician test and push` won't as it has old dependencies.


odc/stats/model.py
```diff
@@ -318,15 +318,8 @@ def _lineage(self) -> Tuple[UUID, ...]:
 
         # TODO: replace this and test
```
Ariana-B: Can we resolve this TODO?

odc/stats/model.py
```diff
@@ -318,15 +318,8 @@ def _lineage(self) -> Tuple[UUID, ...]:
...
-            lineage = tuple(
-                set(
-                    x
-                    for ds in self.datasets
-                    for y in ds.metadata.sources.values()
-                    for x in y.values()
-                )
-            )
+        if "fused" in ds.product.name:
+            lineage = tuple(set(x for ds in self.datasets for x in ds.metadata.sources))
```
Ariana-B: Not sure if I implemented this wrong in the first place or if I'm misremembering now - doesn't the new `ds.metadata.sources` still have the classifiers as the keys? I think this needs to use `ds.metadata.sources.values()`

odc/stats/io.py
```diff
@@ -2,7 +2,13 @@
 Various I/O adaptors
 """
 
-from typing import Any, Dict, List, Optional, Union, cast
+from __future__ import annotations
+
+from typing import Any, Dict, List, Optional, Union, cast, TYPE_CHECKING
```
SpacemanPaul: From Python 3.10, `Dict`, `List`, `Optional` and `Union` should no longer be required.

- For dicts and lists (and sets) simply use the normal lowercase `dict`, `list`, `set`.  e.g. `nums: list[int] = [1,2,3]`
- For `Union`, use the `|` operator.  E.g. `num: float | int`
- For `Optional`, just use `| None`.  E.g. `param: str | None = None`

Not strictly necessary, but I think these changes make code much more readable.

odc/stats/model.py
emmaai: confirmed your implementation is correct, the new `ds.metadata.sources` is a list instead of dict now, e.g., 
```['71cea3a7-e6eb-5a90-af24-2ae9e0ef61e4', '26d96c8d-eae5-5d3d-95f3-361efa6c468a']```

Comment (Ariana-B) [reply odc/stats/model.py]: Good to know my past self knew what she was doing!

Comment (SpacemanPaul) [reply odc/stats/model.py]: Lineage handling changes significantly in the postgis driver, but should be backwards compatible in 1.9 for the postgres driver.

I'm perplexed by this whole section of code though. Why is there magic significance being placed on the presence of substrings in the product name?  Feels very hacky.

File: odc/stats/model.py [hunk 4]
```diff
@@ -316,17 +316,8 @@ def location(self) -> str:
...
-                set(
-                    x
-                    for ds in self.datasets
-                    for y in ds.metadata.sources.values()
-                    for x in y.values()
-                )
-            )
+        if ds.metadata_doc["properties"].get("fused", False):
+            lineage = tuple(set(x for ds in self.datasets for x in ds.metadata.sources))
         else:
```
SpacemanPaul: Further to Ariana's resolved conversation above.

Although lineage handling in the postgres driver is backwards compatible with 1.8, this is NOT true for the postgis driver.

Happy to postpone the postgis handling to a later PR, but can we please wrap this function in a check:

```
if ds.source_tree:
    raise NotImplementError("Postgis driver lineage API not yet supported")
else:
    # Use legacy postgres driver lineage API.
    ....
```

SpacemanPaul (changes requested):
A couple of minor tweaks and/or clarifications requested.  Otherwise looking great!

> tests on the older docker image statistician test and push won't as it has old dependencies.

It should be possible to get this working before merge, and doing so would make me feel a bit more comfortable - we do always run stats in a docker container after all.  Happy to have different docker image tags for 1.8 and 1.9 branches if that helps.

odc/stats/model.py
SpacemanPaul: Also  I'm perplexed by this whole section of code. Why is there magic significance being placed on the presence of substrings in the product name? Feels very hacky.

emmaai:
> A couple of minor tweaks and/or clarifications requested. Otherwise looking great!
> 
> > tests on the older docker image statistician test and push won't as it has old dependencies.
> 
> It should be possible to get this working before merge, and doing so would make me feel a bit more comfortable - we do always run stats in a docker container after all. Happy to have different docker image tags for 1.8 and 1.9 branches if that helps.

It is possible but it'll defy the purpose of failing it in the long run. the docker image won't be published/pushed if failure after merging, which guarantees:
- either `stats` or `dependencies` failure would stop docker image into production
- the dependencies-only update will guarantee old version of stats is till working (hence failed in this pr as old version of stats not working anymore)

odc/stats/model.py
emmaai: It's very hacky because tracting lineage from multiple products within one dataset is never fully supported. `fused` is to indicate that the source datasets are from more than one product. It actually has nothing to do with the postgis driver, the source datasets are cached in a "quasi" metadata form in query step. Here is to extract information and assemble into "real" metadata such that db is not required and decoupled from processing step.

SpacemanPaul:
> > A couple of minor tweaks and/or clarifications requested. Otherwise looking great!
> > > tests on the older docker image statistician test and push won't as it has old dependencies.
> > 
> > 
> > It should be possible to get this working before merge, and doing so would make me feel a bit more comfortable - we do always run stats in a docker container after all. Happy to have different docker image tags for 1.8 and 1.9 branches if that helps.
> 
> It is possible but it'll defy the purpose of failing it in the long run. the docker image won't be published/pushed if failure after merging, which guarantees:
> 
OK I was confused about how the github actions work. Agreed.

odc/stats/model.py
SpacemanPaul: ooh.  You could use `datacube.model.lineage.LineageRelation(s)`?

odc/stats/model.py
```diff
@@ -398,47 +398,30 @@ def render_assembler_metadata(
...
+                    instruments += dataset.metadata_doc["properties"]["eo:instrument"]
+                else:
+                    instruments += [dataset.metadata_doc["properties"]["eo:instrument"]]
+
+            dataset.metadata_doc.setdefault("$schema", "")
+            source_datasetdoc = serialise.from_doc(
+                dataset.metadata_doc, skip_validation=True
+            )
+            # it will grab all useful input dataset preperties
+            dataset_assembler._inherit_properties_from(
```
Ariana-B, line 418: Is this still necessary now that we're no longer passing `dataset_assembler` the dataset doc?

emmaai: the serialised dataset doc is still passed to dataset_assembler to grab some properties. I broke the same functionality in 3 parts, 
- validate classifier `401`
- inherit properties from source `418`
- record lineage `422`

Ariana-B: But doesn't this mean that we're treating the one dataset as both the source the derived dataset? Or does the distinction not particularly matter in this case

odc/stats/model.py
```diff
@@ -313,24 +313,14 @@ def location(self) -> str:
...
-                    for y in ds.metadata.sources.values()
-                    for x in y.values()
-                )
+    def _lineage(self) -> tuple[UUID, ...]:
+        lineage = set()
+        for ds in self.datasets:
+            tree = LineageTree.from_eo3_doc(ds.metadata_doc)
+            lineage |= (
+                tree.child_datasets() if tree.child_datasets() else {tree.dataset_id}
             )
```
SpacemanPaul, line 322: Oh yeah, that's much better.

odc/stats/model.py
SpacemanPaul: It all gets squashed into a set and sorted below, so doesn't matter any more.

SpacemanPaul (approved):

PR #200, opened 2025-07-16
Cross-repo refactor of stac/eo3 code

SpacemanPaul:
ODC-Stats arm of the big eo3/STAC refactor.

- Remove explicit dependency on odc-stac
- pin to core >=1.9.6
- import stac2ds from its new home in core instead of from odc-stac

**See Also**
- https://github.com/opendatacube/datacube-core/pull/2027
- https://github.com/opendatacube/odc-stac/issues/219

### DIFF
--- a/odc/stats/_stac_fetch.py
+++ b/odc/stats/_stac_fetch.py
@@ -1,7 +1,7 @@
 import json
 
 from odc.aio import S3Fetcher, s3_find_glob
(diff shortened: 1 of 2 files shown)


emmaai (approved):

omad:
Updated description with links to the odc-stac PR and the datacube-core PR that necessitated this change.

PR #202, opened 2025-08-08
odc-stats plugin documentation notebook

cbur24:
Notebook and accompanying plugin code and config that shows a simple example of creating a custom odc-stats plugin. The overall goal of the notebook is to provide guidance to EO scientists on how to translate their analysis code to odc-stats.

I wasn't sure on where the notebook and configs should ideally exist in the repo, so I took an educated guess - happy to move it if need be. 

Aims:
* Demystify the use of odc-stats by demonstrating development of a minimal odc-stats plugin.
* Provide example code for running odc-stats functions within a local machine, thus demonstrating how to develop and test the development of an odc-stats "plugin".

This notebook was tested on the following odc libraries, and `datacube==1.9.5`

odc-algo                          1.0.1
odc-apps-cloud                    0.2.3
odc-apps-dc-tools                 1.9.2
odc-cloud                         0.2.5
odc-dscache                       1.9.0
odc-geo                           0.5.0rc1
odc-io                            0.2.2
odc-loader                        0.5.1
odc-stac                          0.4.0
odc-stats                         1.9.0
odc-ui                            0.2.1


odc/stats/plugins/ls_ndvi_median.py
```diff
@@ -0,0 +1,104 @@
...
+    """
+
+    NAME = "ndvi_median"
+    SHORT_NAME = NAME
+    VERSION = "1.0"
+    PRODUCT_FAMILY = "ndvi"
+
+    def __init__(
+        self,
+        input_bands: Sequence[str] = None,
```
pjonsson: I appreciate the type annotations in the example material.

The type annotations that assign None to non-optional types generate complaints by modern type checkers though, for pre-Python 3.10 the type should be `Optional[Sequence[str]]`, and for 3.10 and later one can use `Sequence[str] | None`.

SpacemanPaul (approved):
Hey Chad, thanks so much for this.  This is a great resource.  Very clear and well written - and way better than the nothing we had previously.  We can further refine in future PRs if necessary.

I endorse Peter's comments about ` xxx | None` type hints (datacube 1.9 requires 3.10, so stick with that format).  Otherwise happy to merge.

cbur24:
@SpacemanPaul @pjonsson Thank you for the reviews! I have altered the type annotations to the more modern style.  Just confirming before I merge - are we okay with the location of the notebook under `odc-stats/docs`? Noting I'll also make a seperate PR to link to this notebook in the top-level README.

One of the checks is failing - anything to worry about?

SpacemanPaul:
Yes /docs in the repo is right place for this.

The test failure is nothing you need to worry about just yet - looks like an upstream issue in our build pipeline.  Just hold off merging until we get to the bottom of it.

pjonsson:
> The test failure is nothing you need to worry about just yet - looks like an upstream issue in our build pipeline.

I think it is https://github.com/pypa/setuptools-scm/issues/1216 that is causing the failure. I just removed a bunch of those `Version = attr: ..` in other ODC repositories when I converted them to use `pyproject.toml`, but it seems there are more repositories to convert.

SpacemanPaul:
> > The test failure is nothing you need to worry about just yet - looks like an upstream issue in our build pipeline.
> 
> I think it is [pypa/setuptools-scm#1216](https://github.com/pypa/setuptools-scm/issues/1216) that is causing the failure. 

Yes it's definitely setuptools-scm.  Seem to be some embarassing regression failures happening over there - more than half of the last half dozen or so releases have been yanked.  Ariana has a fix for this repo in #206 

(Made worse by the fact that the CI in this repo doesn't use `build` to build the wheel - #206 will fix that too.)

cbur24:
@Ariana-B Any blockers on merging this one?

PR #205, opened 2025-08-12
CI: use OIDC for CodeCov

pjonsson:
This uses a short-lived token
which is better for security.

Also throw in a commit which pins
the CI action versions.


pjonsson:
The only thing running before the failing step is the checkout action, and while this PR updates that, the checkout doesn't fail, so I'm pretty sure this is a latent failure that isn't seen because the workflow doesn't usually run.

andrewdhicks (approved):
Looks good, I agree that we should be pinning versions a lot more.

PR #206, opened 2025-08-21
Add three-year rolling frequency option

### LINKED ISSUE #203: New frequency option for `save-tasks`
For a summary product that will be built from three-year rolling stats, [cloud_proba_percentiles](https://github.com/opendatacube/odc-stats/blob/develop/odc/stats/plugins/cloud_proba_percentiles.py), I require a multi-year rolling temporal frequency option in `save-tasks`, which is currently not supported. I imagine this option will be analogous to `rolling-3month` option.
- https://github.com/opendatacube/odc-stats/blob/7f34c86bdbd481340c41b5be7e0d0873ce3b3e1c/odc/stats/_cli_save_tasks.py#L24.

The final, indexed product should have the middle year of the three years at the datetime in the metadata.  For example, for a rolling three product where data is summarised from 2019-2021,the metadata should look something like this:

```
properties:
  datetime: 2020-01-01 00:00:00Z
  dea:dataset_maturity: final
  dea:product_maturity: stable
  dtr:end_datetime: 2020-12-31 23:59:59.999999Z
  dtr:start_datetime: 2020-01-01 00:00:00Z
```

The `ga_s2_tidal_composites_cyear_3` product follows this same pattern: `s3://dea-public-data/derivative/ga_s2_tidal_composites_cyear_3/`

Ariana-B:
As per #203

### DIFF
--- a/odc/stats/io.py
+++ b/odc/stats/io.py
@@ -347,13 +347,19 @@ class S3COGSink:
         )  # stac_meta is Python str, but content is 'Dict format'
 
     def dump_with_pystac(
-        self, task: Task, ds: xr.Dataset, aux: xr.Dataset | None = None, proc: StatsPluginInterface = None
+        self,
+        task: Task,
+        ds: xr.Dataset,
+        aux: xr.Dataset | None = None,
+        proc: StatsPluginInterface = None,
     ) -> Delayed:
         """
         Dump files with STAC metadata file, which generated from PySTAC
         """
         json_url = task.metadata_path("absolute", ext=self._stac_meta_ext)
-        meta = task.render_metadata(ext=self._band_ext, use_center_time=hasattr(proc, "CENTER_TIMERANGE"))
+        meta = task.render_metadata(
+            ext=self._band_ext, use_center_time=hasattr(proc, "CENTER_TIMERANGE")
+        )
         json_data = dump_json(meta).encode("utf8")
 
         # fake write result for metadata output, we want metadata file to be
@@ -400,7 +406,9 @@ class S3COGSink:
         sha1_url = task.metadata_path("absolute", ext="sha1")
         proc_info_url = task.metadata_path("absolute", ext=self._proc_info_ext)
         dataset_assembler = task.render_assembler_metadata(
-            ext=self._band_ext, output_dataset=ds, use_center_time=hasattr(proc, "CENTER_TIMERANGE")
+            ext=self._band_ext,
+            output_dataset=ds,
+            use_center_time=hasattr(proc, "CENTER_TIMERANGE"),
         )
 
         dataset_assembler.extend_user_metadata(
--- a/odc/stats/io.py
+++ b/odc/stats/io.py
@@ -347,13 +347,13 @@ class S3COGSink:
         )  # stac_meta is Python str, but content is 'Dict format'
 
     def dump_with_pystac(
-        self, task: Task, ds: Dataset, aux: Dataset | None = None
+        self, task: Task, ds: xr.Dataset, aux: xr.Dataset | None = None, proc: StatsPluginInterface = None
     ) -> Delayed:
         """
         Dump files with STAC metadata file, which generated from PySTAC
         """
         json_url = task.metadata_path("absolute", ext=self._stac_meta_ext)
-        meta = task.render_metadata(ext=self._band_ext)
+        meta = task.render_metadata(ext=self._band_ext, use_center_time=hasattr(proc, "CENTER_TIMERANGE"))
         json_data = dump_json(meta).encode("utf8")
 
         # fake write result for metadata output, we want metadata file to be
@@ -388,8 +388,8 @@ class S3COGSink:
     def dump_with_eodatasets3(
         self,
         task: Task,
-        ds: Dataset,
-        aux: Dataset | None = None,
+        ds: xr.Dataset,
+        aux: xr.Dataset | None = None,
         proc: StatsPluginInterface = None,
     ) -> Delayed:
         """
@@ -400,7 +400,7 @@ class S3COGSink:
         sha1_url = task.metadata_path("absolute", ext="sha1")
         proc_info_url = task.metadata_path("absolute", ext=self._proc_info_ext)
         dataset_assembler = task.render_assembler_metadata(
-            ext=self._band_ext, output_dataset=ds
+            ext=self._band_ext, output_dataset=ds, use_center_time=hasattr(proc, "CENTER_TIMERANGE")
         )
 
         dataset_assembler.extend_user_metadata(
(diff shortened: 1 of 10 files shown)


odc/stats/model.py
```diff
@@ -427,13 +438,19 @@ def render_assembler_metadata(
...
         dataset_assembler.geometry = self.geobox.extent.geom
 
-        dataset_assembler.datetime = format_datetime(self.time_range.start)
-        dataset_assembler.properties["dtr:start_datetime"] = format_datetime(
-            self.time_range.start
-        )
-        dataset_assembler.properties["dtr:end_datetime"] = format_datetime(
-            self.time_range.end
-        )
+        if use_center_time:
```
pjonsson: Introducing a conditionally assigned variable:
```python
time = self.time_range.center if use_center_time else self.time_range
```
allows keeping the previous code without if-statements if the `self.time_range` is replaced with `time` in each RHS.

odc/stats/io.py
```diff
@@ -400,7 +400,7 @@ def dump_with_eodatasets3(
         sha1_url = task.metadata_path("absolute", ext="sha1")
         proc_info_url = task.metadata_path("absolute", ext=self._proc_info_ext)
         dataset_assembler = task.render_assembler_metadata(
-            ext=self._band_ext, output_dataset=ds
+            ext=self._band_ext, output_dataset=ds, use_center_time=hasattr(proc, "CENTER_TIMERANGE")
```
pjonsson: Will setting `CENTER_TIMERANGE = False` in a class work as intended with this logic?

odc/stats/utils.py
```diff
@@ -213,6 +243,34 @@ def mk_rolling_season_rules(temporal_range, months, interval):
     return rules
 
 
+def mk_rolling_years_rules(temporal_range, years, interval):
```
pjonsson: Can this new function get type annotations?

odc/stats/utils.py
```diff
@@ -132,6 +132,36 @@ def bin_annual(
...
+    years: int,
+    interval: int,
+) -> dict[tuple[str, int, int], list[CompressedDataset]]:
+    binner = rolling_season_binner(
+        mk_rolling_years_rules(temporal_range, years, interval)
+    )
+
+    tasks = {}
+    for tidx, cell in cells.items():
+        # This is a great pylint warning, but doesn't apply here because we
```
pjonsson: Can't this entire warning+suppression be avoided by removing `utc_offset` and putting `cell.utc_offset` inside the lambda?

odc/stats/utils.py
```diff
@@ -132,6 +132,36 @@ def bin_annual(
...
+        utc_offset = cell.utc_offset
+        _grouped = toolz.groupby(lambda ds: binner(ds.time + utc_offset), cell.dss)
+
+        grouped = defaultdict(list)
+        for key, value in _grouped.items():
+            for k in key:
+                grouped[k].extend(value)
+
+        for temporal_k, dss in grouped.items():
+            if temporal_k != "":
```
pjonsson: It doesn't make a huge difference with simplifying the logic in this case, but if you move this if-statement to the insertion loop instead it will at least save some work/memory allocations.

odc/stats/io.py
```diff
@@ -347,13 +347,19 @@ def get_eo3_stac_meta(
         )  # stac_meta is Python str, but content is 'Dict format'
 
     def dump_with_pystac(
-        self, task: Task, ds: Dataset, aux: Dataset | None = None
+        self,
+        task: Task,
+        ds: xr.Dataset,
+        aux: xr.Dataset | None = None,
+        proc: StatsPluginInterface = None,
```
pjonsson: I somehow managed to comment on one of the intermediate commits, second attempt:

`StatsPluginInterface` is a class, so that can't be None.

odc/stats/model.py
```diff
@@ -493,8 +517,13 @@ def render_metadata(
 
         properties: dict[str, Any] = deepcopy(product.properties)
 
-        properties["dtr:start_datetime"] = format_datetime(self.time_range.start)
-        properties["dtr:end_datetime"] = format_datetime(self.time_range.end)
+        if use_center_time:
```
pjonsson: Same comment here about introducing an intermediate variable for the time and use that in the RHS instead of the if-statement.

odc/stats/utils.py
```diff
@@ -132,6 +132,36 @@ def bin_annual(
     return tasks
 
 
+def bin_rolling_annual(
+    cells: dict[tuple[int, int], Cell],
+    temporal_range,
```
pjonsson: Can `temporal_range` get a type annotation too?

odc/stats/utils.py
```diff
@@ -132,6 +132,36 @@ def bin_annual(
...
+    tasks = {}
+    for tidx, cell in cells.items():
+        # This is a great pylint warning, but doesn't apply here because we
+        # only call the lambda from inside each iteration of the loop
+        # pylint:disable=cell-var-from-loop
+        utc_offset = cell.utc_offset
+        _grouped = toolz.groupby(lambda ds: binner(ds.time + utc_offset), cell.dss)
+
+        grouped = defaultdict(list)
+        for key, value in _grouped.items():
```
pjonsson: There's a loop over `key` in the next line, is there a single key, or multiple keys?

odc/stats/io.py
Ariana-B: The idea is that the `CENTER_TIMERANGE` attribute would only be added to classes that need it. At this time, I don't see a use case for its value to matter.
@cbur24 thoughts?

odc/stats/utils.py
Ariana-B: Each key is a tuple (see `tests/test_utils.py::test_rolling_season_binner`)

odc/stats/io.py
pjonsson: I have to admit I would be a bit surprised if I set `CENTER_TIMERANGE = False` in a class and that gave me center time.

I'm not the target user for this though, so that probably doesn't matter. But how are classes that don't want center time but inherit from a class that sets `CENTER_TIMERANGE` meant to deal with their situation?

cbur24 (approved):
Speaking only from a user testing point-of-view, I can confirm that these code changes result in the correct outputs. For a `save-tasks` run with `--temporal-range 2020--P3Y` and `--frequency rolling-3years`, the datetime metadata on the results are:
```
properties:
  datetime: 2021-01-01 00:00:00Z
  dea:dataset_maturity: final
  dtr:end_datetime: 2021-12-31 23:59:59.999999Z
  dtr:start_datetime: 2021-01-01 00:00:00Z
```
Meaning the center year is correctly identified as the labelling year.  I tested using the following odc libraries (with python 3.10.18):
```
odc-algo==1.0.1
odc-dscache==1.9.1
odc-stats==1.9.3.dev11+gdeecab5
datacube==1.9.7
```
I can also confirm that the geotiff results are identical if I run the same code but with `--frequency all` over the same temporal range.

SpacemanPaul (approved):
As discussed I think we will have to revisit this at some point.

I think in the longer term we will need to have two separate date-time range fields in metadata:

1) A non-overlapping range for the middle period used for searching and descriptive purporses
2) An overlapping range for the source data contributing to the full rolling window.

But that needs further discussion, I think this approach is good for the immediate/short term.

PR #207, opened 2025-09-05
Use geomad instead of hdstats.

SpacemanPaul:
Use new open source [Geomad](https://github.com/GeoscienceAustralia/geomad) package instead of hdstats.

Remove dependency on `packages.dea.ga.gov.au` python package index.


omad (approved):
Nifty!

SpacemanPaul:
Tests are hanging due to worker memory blowing out.  Not sure what the issue is at this stage.

omad:
That memory limit argument is used to tell dask how much memory it is allowed to use (and maybe to work out chuck sizes, I'm not sure).

I think GitHub actions runners have 16GB of RAM, so dask needs to use substantially less to not be killed.

The log messages from dask in the failed runs indicated it had decided to use about 14GB, so it must use a heuristic that assumes it can use most of the available ram, too much in this case.

SpacemanPaul:
Comments:

I'm having to leave this where it is for now, as I'm off on a much-needed 4 week holiday.

Status

- Geomad seems to be working hunky dory in odc-algo.
- integration test failures here appear to be happening BEFORE the execution gets to the geomad code.  I think it's got something to do with the dask scheduling of the output COG writing.
- integration tests have been running against the last released version of odc-stats code (i.e. the code from the previous successfully merged PR), rather than the code of the current PR, making it very difficult to figure out where things broke.  I have fixed this in this PR.
- I have reverted to core's internal (deprecated) `to_cog()` method which gives clearer error messages (odc-geo `to_cog` just runs out of memory and the container gets killed).  With datacube internal `to_cog()` I get pickling errors referencing `_HLGExprSequence` which I have seen before and indicates that an internal Dask object (e.g. a `Client`) is getting passed to a delayed dask function, but I haven't been able to spot where this might be happening.

There are some unrelated test errors happening due to the switch to numpy 2.  These should be relatively straightforward to resolve.

omad:
In addition to @SpacemanPaul 's suspicion about the different `to_cog()` dask functions, I'm wondering whether the `yxbt_sink` and `reshape_yxbt` features in odc-stats are still necessary.

My understanding is that they were very important a few years ago to efficiently scale across dask workers, especially for running all of time statistical summaries. Those summaries are embarrassingly parallel spatially, but, (I think), the dask scheduler at the time was causing major problems in how it was splitting the work up.

In the last year there have been a few significant improvements in the dask/distributed scheduler that should address some or all of these issues, and the custom memory sinks may just be getting in the way.

It may be possible to use a much simpler implementation using the geomad kernel with xarray code something like the following that ran better against one of these new tests:

```
def xr_geomedian(x, dim="time"):
    return xr.apply_ufunc(
        geomad.nangeomedian_pcm,
        x,
        input_core_dims=[["time"]],
        dask="parallelized",
        dask_gufunc_kwargs={'allow_rechunk': True}
        )
src_dataarray = src.to_dataarray(dim="band").chunk({'x': 800, 'y': 800, 'time': -1, 'band': -1})
result = xr_geomedian(src_dataarray)
```

It also shouldn't be necessary to reach as low level as the `.map_blocks()` calls, since `.apply_ufunc()` can do simple dimension reductions like geomad on it's own.

## Loading
I'm also curious as to whether there are custom dask loading functions implemented here in `odc-stats` that would be better replaced by the code in `odc-loader`.

cbur24:
@SpacemanPaul @omad 
I ran Landsat 8 geomedians using the code in this branch to see if the memory leak issue you're reporting is replicated outside of the unit testing. On the `unstable` sandbox image, the landsat geomedian plugin (`gm-ls`) worked as expected without memory spikes.  The outputs also look correct. Maybe this is only an issue with the unit tests and not with the geomedian code per se?

I've pushed the jupyter notebook and config file that I used for this testing into `odc-stats/docs` within this branch so you can test for yourselves - we can delete these later once your happy. ODC and Datacube versions tested:

UPDATE: this was run on `xarray=2025.6.1`

image

docker/requirements.txt
```diff
@@ -1,16 +1,14 @@
...
+datacube[performance,s3]>=1.9.5
 eodatasets3>1.9
-hdstats==0.1.8.post1
 numexpr>=2.11
-odc-algo>=1.0.1,<1.1 # Don't want geomad yet
+odc-algo>=1.1.1
 odc-apps-cloud>=0.2.2
 # For testing
-odc-apps-dc-tools>=0.2.12
+odc-apps-dc-tools>=0.2.12,<1.9.3 # 1.9.3 requires datacube 1.9.6
```
pjonsson: Line 5 permits 1.9.6, do you still want this limitation?

docker/requirements.txt
```diff
@@ -1,16 +1,14 @@
 
---extra-index-url https://packages.dea.ga.gov.au/
 # For ML
 ai-edge-litert
-datacube-ows>=1.9,<1.9.8 # 1.9.8 drops py3.10
-datacube[performance,s3]==1.9.5
+datacube-ows[webserver]>=1.9,<1.9.8 # 1.9.8 drops py3.10
+datacube[performance,s3]>=1.9.5
```
pjonsson: Should this also be relaxed in `pyproject.toml`?

docs/ga_ls8cls9c_gm_cyear_3.yaml
```diff
@@ -0,0 +1,95 @@
...
+        nbart_red: 1
+      green:
+        nbart_green: 1
+      blue:
+        nbart_blue: 1
+    scale_range:
+    - 0
+    - 3000
+
+aws_unsigned: True
```
SpacemanPaul: This should be "true" (all lower case) as per pre-commit warning.

docs/ga_ls8cls9c_gm_cyear_3.yaml
```diff
@@ -0,0 +1,95 @@
...
+
+max_processing_time: 2400
+job_queue_max_lease: 300
+renew_safety_margin: 60
+future_poll_interval: 2
+# Generic product attributes
+cog_opts:
+  zlevel: 9
+apply_eodatasets3: True
```
SpacemanPaul: And this one too.

tests/test_gm_ls.py
```diff
@@ -177,6 +177,7 @@ def test_resampling(dataset):
 
 
 def test_no_data_value(monkeypatch):
+    _ = pytest.importorskip("hdstats")
```
SpacemanPaul: Why is this test `importorskip`ing `hdstats` ?  Shouldn't this be `geomad` now?

pyproject.toml
```diff
@@ -5,25 +5,32 @@ description = "Statistical Product Generation Framework"
...
+requires-python = ">=3.12"
 readme = "README.md"
 license = "Apache-2.0"
+classifiers = [
+    "Programming Language :: Python :: 3",
+    "Programming Language :: Python :: 3.12",
+    "Programming Language :: Python :: 3.13",
+]
 dependencies = [
+    "affine<3",
```
SpacemanPaul, line 17: Oo.  Affine 3 is out and has some breaking API changes.  I wonder if that will affect any other ODC repos...

pyproject.toml
```diff
@@ -36,21 +43,40 @@ dependencies = [
...
     "pytest",
     "pytest_httpserver",
-    "moto"
+    "moto",
+    "odc-apps-dc-tools",
 ]
 ows = [
     "blinker",
-    "datacube_ows>=1.9",
+    "datacube_ows[webserver]>=1.9", # need matplotlib for thumbnails
```
SpacemanPaul, line 66: We shouldn't need `[webserver]` ....  That will need to be fixed in ows though,  (We need a `[stylingapi`] optional dependency group.)

SpacemanPaul (commented):
A few minor comments - not all actionable.

I can't actually approve because I originally created the PR, but I approve in spirit.

pyproject.toml
pjonsson: I know it affects the eo-datasets repository, not sure about any others.

PR #208, opened 2025-09-05
Replace tflite with ai-edge-litert

omad:
TFLite passed away a long time ago, it rose again as `ai-edge-litert`

https://ai.google.dev/edge/litert/migration#migrate-python

### DIFF
--- a/docker/requirements.txt
+++ b/docker/requirements.txt
@@ -1,11 +1,13 @@
 
 --extra-index-url https://packages.dea.ga.gov.au/
+# For ML
+ai-edge-litert
 datacube-ows>=1.9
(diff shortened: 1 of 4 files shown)


omad:
Updated to *only* switch from `tflite-runtime` to `ai-edge-litert`. And not try to replace `hdstats` with `geomad` as well, waaaay too ambitious.

Ariana-B (approved):
Might also be worth renaming `TensorFlowLiteModelPlugin`?

omad:
> Might also be worth renaming `TensorFlowLiteModelPlugin`?

For sure! But not for me this week sorry.

PR #210, opened 2025-09-07
With Geomad and LESS ram for integration tests

omad:
See #207, this is that, but with lowered RAM usage allowed in GHA.

From my comment https://github.com/opendatacube/odc-stats/pull/207#issuecomment-3257350707
> That memory limit argument is used to tell dask how much memory it is allowed to use (and maybe to work out chuck sizes, I'm not sure).
> 
> I think GitHub actions runners have 16GB of RAM, so dask needs to use substantially less to not be killed.
> 
> The log messages from dask in the failed runs indicated it had decided to use about 14GB, so it must use a heuristic that assumes it can use most of the available ram, too much in this case.

### DIFF
--- a/docker/requirements.txt
+++ b/docker/requirements.txt
@@ -1,11 +1,10 @@
 
---extra-index-url https://packages.dea.ga.gov.au/
 datacube-ows>=1.9
 datacube[performance,s3]>=1.9.5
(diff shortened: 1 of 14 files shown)


SpacemanPaul:
🤞

PR #212, opened 2025-09-08
docker: pull quietly

pjonsson:
The progress meter produces a lot
of spam in the CI logs, so disable
it.


pjonsson:
Looking at the CI logs with all the spam is quite annoying, so merging this.

PR #213, opened 2025-09-08
CI: update+pin actions by hash

pjonsson:


pjonsson:
Apparently I already did this in #205, but it's not merged.

PR #217, opened 2025-10-07
Allow dask to spill to disk

andrewdhicks:
The integration tests are failing due to high memory usage. 
This change 
* reduces the spatial size of the array to handle the year-long time depth.
* allows dask distributed to spill to disk when memory usage is high.

### DIFF
--- a/docker/distributed.yaml
+++ b/docker/distributed.yaml
@@ -2,7 +2,7 @@
 distributed:
   worker:
     memory:
-      target: false
-      spill: false
+      target: 0.65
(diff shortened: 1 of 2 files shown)


andrewdhicks:
Ah, currently geomedian is hardcoded to use the `mem` strategy, which cannot be spilled to disk.
And the unit tests are failing, on top of the integration tests.

andrewdhicks:
Closing, as this approach looks like it doesn't resolve the issue

PR #218, opened 2025-10-13
GM memory limit

andrewdhicks:
To fix the failing integration test for geomedian:
Reduce memory limit to 12G
increase threads to 4

### DIFF
--- a/.github/workflows/statistician-image.yml
+++ b/.github/workflows/statistician-image.yml
@@ -8,6 +8,7 @@ on:
     paths:
       - 'docker/**'
       - '!docker/readme.md'
+      - 'tests/integration_test.sh'
   push:
     branches:
       - develop
--- a/docker/distributed.yaml
+++ b/docker/distributed.yaml
@@ -1,8 +1,13 @@
 # Note that this file is used by Dask configuration
 distributed:
+  scheduler:
+    active-memory-manager:  # See https://distributed.dask.org/en/stable/worker-memory.html#ignore-process-memory
+      measure: managed
   worker:
     memory:
+      rebalance:
+        measure: managed
       target: false
       spill: false
-      pause: 0.95
-      terminate: 0.97
+      pause: false
+      terminate: false
--- a/docker/distributed.yaml
+++ b/docker/distributed.yaml
@@ -1,5 +1,8 @@
 # Note that this file is used by Dask configuration
 distributed:
+  nanny:
+    pre-spawn-environ:
+      MALLOC_TRIM_THRESHOLD_: 0
   scheduler:
     active-memory-manager:  # See https://distributed.dask.org/en/stable/worker-memory.html#ignore-process-memory
       measure: managed
--- a/tests/integration_test.sh
+++ b/tests/integration_test.sh
@@ -12,7 +12,7 @@ odc-stats --version
 echo "Test LS GeoMAD"
 
 odc-stats save-tasks --config https://raw.githubusercontent.com/GeoscienceAustralia/dea-config/709daaee176c04e33de4cc9600462717cca5b34d/dev/services/odc-stats/geomedian/ga_ls8c_nbart_gm_cyear_3.yaml --year=2015 --tiles 49:50,24:25 --overwrite ls-geomad-cyear.db
-odc-stats run  --threads=1 --config https://raw.githubusercontent.com/GeoscienceAustralia/dea-config/709daaee176c04e33de4cc9600462717cca5b34d/dev/services/odc-stats/geomedian/ga_ls8c_nbart_gm_cyear_3.yaml --location file:///tmp --overwrite ls-geomad-cyear.db
+odc-stats run  --memory-limit=12GiB --threads=4 --config https://raw.githubusercontent.com/GeoscienceAustralia/dea-config/709daaee176c04e33de4cc9600462717cca5b34d/dev/services/odc-stats/geomedian/ga_ls8c_nbart_gm_cyear_3.yaml --location file:///tmp --overwrite ls-geomad-cyear.db
 
 ./tests/compare_data.sh /tmp/x49/y24/ ga_ls8c_nbart_gm_cyear_3_x49y24_2015--P1Y_final*.tif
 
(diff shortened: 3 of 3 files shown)


omad:
There's a setting for the dask scheduler to disable this behaviour. It's not a useful thing for CI. 😀

`2025-10-13 06:09:30,358 - distributed.worker.memory - WARNING - Worker is at 95% memory usage. Pausing worker.  Process memory: 11.48 GiB -- Worker memory limit: 12.00 GiB`

I was trying to see which version of which library caused this memory explosion, but wasn't able to get an answer.

andrewdhicks:
I've been able to run the GM test on the sandbox with the same versions (I think) with a 12G limit, and it never went above 10G.
Possibly a difference in the way the OS or docker free up memory?
This test looks to have started failing when the code was updated to datacube 1.9, so there are a lot of potential causes.

I've got a version of the code to configure the geomedian reshape_strategy, which would allow `yxbt` mode, and we could turn spill to disk on. (Mem fails if spilled). Turning off the 'pause' memory threshold would make it fail faster, instead of hanging for an hour.

I'm currently thinking about splitting out the docker build and push (for internal GA use) into a separate repo, and making the GM integration test run on a smaller set of data.

pjonsson:
> I've been able to run the GM test on the sandbox with the same versions (I think) with a 12G limit, and it never went above 10G. Possibly a difference in the way the OS or docker free up memory?

Were you running a docker on top of Linux in the same way as the tests do, and it never went above 10G?

I don't know anything about odc-stats, but unless there is a memory leak in some library, it's usually (g)libc that hangs on to memory after it has been freed. It might be possible to tune `M_TRIM_THRESHOLD` to make glibc more eager to return some memory (manual page for mallopt has more information). Here's a Stackoverflow answer that gives some more details: https://stackoverflow.com/a/48652734

I have no idea if it is actionable, but keep in mind that scheduling order for tasks/threads can affect the maximum amount of live memory.

> Turning off the 'pause' memory threshold would make it fail faster, instead of hanging for an hour.

I've set `timeout-minutes` on most ODC-workflows, so things are killed instead of hanging indefinitely. If the tests should take 10 minutes in this repository it might be worth setting the timeout to 30 minutes so the hanging things are killed and the CI runner becomes available for use by other ODC repositories.

(I think the timeout should be set to something like 2-3x the expected runtime so it's never triggered under usual circumstances. The default timeout is really high so if you manage to hang things so the cancel button doesn't work it will be a long wait.)

> I'm currently thinking about splitting out the docker build and push (for internal GA use) into a separate repo, and making the GM integration test run on a smaller set of data.

We don't use odc-stats currently, but I'm not so keen on having even more repositories to maintain and keep track of. And if we were to use odc-stats, we would be using the docker image, which I imagine is representative for a large share of ODC users.

pjonsson:
From what I can tell, the Docker image in this repository doesn't use the code in the repository, it installs the latest odc-stats from PyPi for some reason.

I have shifted things around so the image installs the current code in the repository and tests that instead in #220.

> I've been able to run the GM test on the sandbox with the same versions (I think) with a 12G limit, and it never went above 10G.

I'm not able to reproduce that, running
```shell
docker compose up -d --build
docker compose exec -e AWS_DEFAULT_REGION=ap-southeast-2 -T stats ./tests/init_db.sh
docker compose exec -T stats  ./tests/integration_test.sh
```
on top of #220 balloons the memory to 30+ GB rather quickly on my machine (regular Ryzen desktop PC with Ubuntu 24.04). I set a memory limit to 32 GiB for the first test, and after waiting a bit more than 9 minutes on the first test, I gave up waiting. The test is using CPU, but I have to finish some other stuff I'm working on.

I tried 4 threads instead of 1 thread, and that put the threads into pause in no time at all, so I suspect you want to make the CI green first, and then make it go faster.

andrewdhicks:
Thanks for running those benchmarks @pjonsson!
There are some similar glibc-related dask-specific things I'll try here: https://distributed.dask.org/en/stable/worker-memory.html#memory-not-released-back-to-the-os
The 10GiB max behaviour was running the code directly, not running within docker, as I wanted to rule out a change to the algorithm as the cause.

The split I'm proposing would move GA's data checking to a private CI pipeline where we can perform a wider range of regression checks on a better resourced runner.  This would leave this GitLab pipeline to do a smaller scale job that can run in the public runner. This fits with the work that @omad has started to make the repo less GA-specific, by moving the test config files into the repo in #210.
Right now the docker image gets pushed to a GA ECR, rather than somewhere public.

pjonsson:
Oh, your link for memory-related things is much more comprehensive than what I linked, and is also targeted at Dask specifically. Yeah, ignore what I said and follow that instead.

> The 10GiB max behaviour was running the code directly, not running within docker, as I wanted to rule out a change to the algorithm as the cause.

Since you've run the test case on your machine with the same dependencies, I'm guessing there is something funny going on with the Docker image since that can't complete the first test with 16 GB of memory. I don't know the history for why the image is constructed the way it is, but it is an interesting choice to install the released version of odc-stats instead of the code currently in this repository, so there might be more interesting choices made with how the image is built.

I also just realized that there is one more thing that might have changed since this repository last had a green CI:  @omad recently pointed out to me that the Github CI runners have 4 cores, but they used to have 2 cores, and I'm guessing Github also doubled the memory when they doubled the number of cores. I don't know anything about Dask or the predecessor of ai-edge-litert, but it's not inconceivable that some dependency of odc-stats would adjust its runtime behavior based on the detected amount of cores/memory.
 
> The split I'm proposing would move GA's data checking to a private CI pipeline where we can perform a wider range of regression checks on a better resourced runner. This would leave this GitLab pipeline to do a smaller scale job that can run in the public runner.

My apologies, I thought you suggested creating another repository under the opendatacube organization when I made my previous comment, what repositories you create elsewhere is none of my business.

I haven't followed the this repository closely, but my impression is that the current tests used to run fine with Datacube 1.8, and after updating to 1.9 and changing a lot of dependencies, the same tests run out of memory by quite a wide margin. It's probably not your problem, but this kind of "regression" with using a lot more resources to process the same thing should be noticeable on your cloud costs.

PR #219, opened 2025-10-15
Pass reshape_strategy to geomedian plugin

andrewdhicks:
Adds geomedian plugin config:

`reshape_strategy: 'yxbt' or 'mem'`

### DIFF
--- a/odc/stats/plugins/gm.py
+++ b/odc/stats/plugins/gm.py
@@ -30,7 +30,7 @@ class StatsGM(StatsPluginInterface):
         basis_band=None,
         aux_names: dict[str, str] = None,
         work_chunks: tuple[int, int] = (400, 400),
-        reshape_strategy : str = "mem",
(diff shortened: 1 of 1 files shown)


odc/stats/plugins/gm.py
```diff
@@ -30,6 +30,7 @@ def __init__(
         basis_band=None,
         aux_names: dict[str, str] = None,
         work_chunks: tuple[int, int] = (400, 400),
+        reshape_strategy: str = "mem",
```
pjonsson, line 33: If you change the type from `str` to `Literal["mem", "yxbt", "string1", "string2"]`, it should make it harder for the caller to pass in bad values.

andrewdhicks: I'd prefer not to define them here so they don't get out of sync but importing the type from the algo package would work.

Ariana-B (approved):

PR #220, opened 2025-10-15
docker: install odc-stats from repository

pjonsson:


andrewdhicks (approved):
Thanks so much for working on this @pjonsson 
This makes much more sense!

PR #224, opened 2025-11-10
setup.cfg: use datacube < 1.10.0

pjonsson:
This will keep the package working
even after the next API breaking
datacube version is released.


alexgleith (approved):
Thanks

PR #226, opened 2025-11-10
Rollback datacube to 1.9.5

andrewdhicks:
Newer versions of datacube are causing issues with odc-stats running with a much larger memory footprint and blocking threads on masking operations.

Datacube 1.9.5 does not have these issues.

(1.9.6+1.9.7: has differences in handling dimensions required for chunking.)

Will roll forward when the issues have been resolved.


pjonsson (approved):

pjonsson:
@andrewdhicks does this make the CI green from now on?

andrewdhicks:
> @andrewdhicks does this make the CI green from now on?

@pjonsson I think once https://github.com/opendatacube/odc-stats/pull/228 goes through, it should all be green!

PR #228, opened 2025-11-12
Update deps for codecov

andrewdhicks:
Newly pinned codecov seems to need more dependencies installed


pjonsson (approved):
@andrewdhicks I haven't seen this problem before, and I did spend a fair amount of time on looking at CodeCov earlier this year when it wasn't working in a lot of ODC repositories. Very strange.

PR #229, opened 2025-11-15
CI: use anchors for paths

pjonsson:
The path lists are meant to be identical,
so use the new yaml anchors to ensure that.


pjonsson:
CI failure is unrelated to the PR and is caused by me making the PR from my forked repository instead of from a branch in the repository under the opendatacube organization.

PR #231, opened 2025-12-03
CICD: Don't publish to S3

andrewdhicks:
The DEA package repository is being decommissioned.
PyPI should be used instead.


pjonsson (approved):
I think the CI failure is caused by a missing upper bound on odc-stac.

This change looks good to me though.

PR #233, opened 2026-01-05
Ensure that the transform is a list, so it can be converted to JSON

alexgleith:
I've been getting an error:

> TypeError: Object of type Affine is not JSON serializable

using the `develop` version of `odc-stats` and `odc-geo==0.5.0`.

I have a simple fix, which is hopefully backwards compatible... no test, though.


alexgleith:
Error in tests is unrelated...

PR #235, opened 2026-01-20
Fix multi-crs concatination bug

andrewdhicks:
Xarray has modified the way dims are exposed from `xr.Dataset`s, which has broken stats when trying to combine datasets of multiple CRSs.
To get the dims in order, we need to access the dims of a `xr.DataArray`.

### DIFF
--- a/odc/stats/io.py
+++ b/odc/stats/io.py
@@ -826,7 +826,9 @@ def load_with_native_transform(
     if len(_xx) == 1:
         xx = _xx[0]
     else:
-        xx = xr.concat(_xx, _xx[0].dims[0])  # type: ignore
+        sample_band = basis if basis is not None else bands[0]
+        concat_dim = _xx[0][sample_band].dims[0]
+        xx = xr.concat(_xx, dim=concat_dim)  # type: ignore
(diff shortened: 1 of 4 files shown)


odc/stats/io.py
```diff
@@ -826,7 +826,9 @@ def load_with_native_transform(
     if len(_xx) == 1:
         xx = _xx[0]
     else:
-        xx = xr.concat(_xx, _xx[0].dims[0])  # type: ignore
+        sample_band = basis if basis is not None else bands[0]
+        concat_dim = _xx[0][sample_band].dims[0]
+        xx = xr.concat(_xx, dim=concat_dim)  # type: ignore
```
pjonsson: Since you're changing this line: can you please change the ignore into `ignore[assign]` so we just ignore the specific issue instead of anything type-related?

And I'm guessing there's a typo on "concatenation" in the PR title.

andrewdhicks: Done! Good pick up.

alexgleith (approved):
Looks good!

We bumped into this error too. Thanks Andrew!

PR #237, opened 2026-02-04
Fix stale CRS metadata after reprojection

taufik-shf:
## Problem

While processing data with odc-stats, I noticed the pipeline correctly reprojects data to the target grid, but the CRS metadata revert to the native CRS downstream. This causes:

- Inconsistent STAC vs GeoTIFF metadata (STAC shows target CRS, GeoTIFF tags show native CRS)
- Downstream tools reading incorrect CRS from file headers

## Root Cause Analysis

> [!IMPORTANT]
> **Custom instrumentation for diagnosis**
> To diagnose this issue, I added **custom CRS/grid logging** at key pipeline points to track where CRS metadata diverges from actual coordinates. All log lines below (marked WARNING - CRS ...) are from this custom instrumentatio. They're not standard odc-stats output. 
>
> The logs track:
> - CRS before/after reprojection  
> - Coordinate ranges (to verify grid transformation)
> - CRS metadata at reducer entry/exit

### Initial observation
bugs_evidence

The logs reveal the inconsistency:

- `pre_reproject` shows native UTM coordinates: x=[315375..373265], y=[43435..-95] -> EPSG:32649
- `post_reproject` shows **coordinates transformed** to the target grid: x=[1.055e+07..1.06e+07], y=[49995..5] -> EPSG:6933 range
- BUT the CRS metadata still reports crs=EPSG:32649

> [!Note]
> **Identified Issue:** The reprojection warps the grid correctly, but the CRS metadata remains stale and affects the downstream.

### Fix 1: Assign CRS after reprojection

- After `xr_reproject`, I explicitly assign the CRS to match the destination grid

- This ensures the post-reproject dataset is internally consistent: both coordinates AND CRS metadata reflect the target grid.

fix_01_bugs_02_evidence_2

- After applying Fix 1:

     - `post_reproject` now correctly shows `crs=EPSG:6933` 
     - Coordinates and CRS metadata are properly aligned
     - **However**, there's still an issue downstream.

> [!Note]
> **Issue:** Downstream in the gm plugin - reducer(), the dataset enters with the grid CRS (EPSG:6933) but exits with the native CRS again (EPSG:32649).

### Fix 2: Prevent CRS regression in gm reducer

**The fix:**
1. Capture the authoritative CRS from `.odc.crs`
2. Remove all stale CRS/grid-mapping metadata
3. Cleanly re-attach the CRS

**Evidence**

bug_fix_evidence

### With both fixes applied:

- [x] Post-reproject CRS is correct 
- [x] Coordinates match the target grid
- [x] Reducer maintains CRS throughout
- [x] No CRS regression
- [x] STAC and GeoTIFF metadata remain consistent

### Other smaller changes

- **Ensure `proj:transform` is JSON serialisable**  
  changed: `transform=geobox.transform` -> `transform=list(geobox.transform)`

I hit a case where `geobox.transform` was still an `Affine` object, which is not JSON serialisable and can break STAC/item serialisation. Converting to `list(...)` ensures we always write a JSON-safe transform.

- **Relax datacube pin and lock odc-stac version**

Changed: `datacube==1.9.5` to `datacube>=1.9.5` and pinned `odc-stac==0.4.0`.

`_stac_fetch.py` imports `stac2ds` from odc-stac. In newer datacube versions (≥1.9.6), stac2ds also exists in datacube.metadata, but as long as we pin odc-stac==0.4.0, the import remains unambiguous and works correctly. 

Tested successfully with datacube 1.9.10.

### DIFF
--- a/odc/stats/io.py
+++ b/odc/stats/io.py
@@ -33,7 +33,7 @@ from pyproj import aoi, transformer
 
 from odc.geo.geobox import GeoBox
 from odc.geo.geobox import pad as gbox_pad
-from odc.geo.xr import xr_reproject
+from odc.geo.xr import xr_reproject, assign_crs
 
 from ._grouper import group_by_nothing, solar_offset
 from odc.algo._masking import (
@@ -813,6 +813,9 @@ def load_with_native_transform(
             chunks=_chunks,
             **extra_args,
         )
+
+        # Ensure output advertises the destination CRS consistently
+        _yy = assign_crs(_yy, crs=geobox.crs)
 
         if isinstance(_yy, xr.DataArray) and vars_to_scale:
             _yy = _yy > 64
--- a/odc/stats/model.py
+++ b/odc/stats/model.py
@@ -528,8 +528,11 @@ class Task:
         )
         ProjectionExtension.add_to(item)
         proj_ext = ProjectionExtension.ext(item)
+
         proj_ext.apply(
-            epsg=geobox.crs.epsg, transform=geobox.transform, shape=list(geobox.shape)
+            epsg=geobox.crs.epsg, 
+            transform=list(geobox.transform),  # ensure transform is JSON serializable
+            shape=list(geobox.shape)
         )
 
         # Lineage last
--- a/odc/stats/model.py
+++ b/odc/stats/model.py
@@ -485,7 +485,7 @@ class Task:
 
         return dataset_assembler
 
-    def render_metadata(
+    def render_metadata(  # pylint: disable=too-many-locals
         self,
         ext: str = EXT_TIFF,
         processing_dt: datetime | None = None,
@@ -530,9 +530,9 @@ class Task:
         proj_ext = ProjectionExtension.ext(item)
 
         proj_ext.apply(
-            epsg=geobox.crs.epsg, 
+            epsg=geobox.crs.epsg,
             transform=list(geobox.transform),  # ensure transform is JSON serializable
-            shape=list(geobox.shape)
+            shape=list(geobox.shape),
         )
 
         # Lineage last
--- a/odc/stats/plugins/gm.py
+++ b/odc/stats/plugins/gm.py
@@ -8,6 +8,7 @@ from odc.algo import geomedian_with_mads
 from ._registry import StatsPluginInterface, register
 from odc.algo import enum_to_bool, erase_bad
 from odc.algo import mask_cleanup
+from odc.geo.xr import assign_crs
 import logging
 
 _log = logging.getLogger(__name__)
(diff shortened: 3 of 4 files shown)


odc/stats/model.py
```diff
@@ -485,7 +485,7 @@ def render_assembler_metadata(
 
         return dataset_assembler
 
-    def render_metadata(
+    def render_metadata(  # pylint: disable=too-many-locals
```
pjonsson, line 488: The change doesn't add any local variables, is this needed?

setup.cfg
```diff
@@ -22,7 +22,8 @@ install_requires =
     botocore
     click>=8.0.0
     dask
-    datacube==1.9.5
+    datacube>=1.9.5
```
pjonsson, line 25: The datacube version is pinned because memory consumption balloons for odc-stats with later versions. I'm not involved in odc-stats so I don't know about the timeline for resolving that.

odc/stats/io.py
```diff
@@ -814,6 +814,9 @@ def load_with_native_transform(
             **extra_args,
         )
 
+        # Ensure output advertises the destination CRS consistently
```
pjonsson, line 817: Shouldn't the output from `xr_reproject` have the right CRS, so this issue should be fixed inside that function instead?

odc/stats/plugins/gm.py
```diff
@@ -123,6 +124,21 @@ def reduce(self, xx: xr.Dataset) -> xr.Dataset:
         gm = geomedian_with_mads(xx, **cfg)
         gm = gm.rename(self._renames)
 
+        # geomedian_with_mads drops spatial_ref; re-attach from input
```
pjonsson, line 127: Same question here, shouldn't `geomedian_with_mads` return an object with the right metadata?

odc/stats/model.py
taufik-shf: Yes, the pylint disable is needed. Running pre-commit gives:

image

Pre-commit checks the whole function since I modified a line in it. Options are: keep the disable, refactor the function, or ignore the check. Happy to revert if you prefer a different approach!

setup.cfg
taufik-shf: Thanks for the context! I wasn't aware of the memory issue with later datacube versions.

I made this change because my environment uses datacube 1.9.10, and the strict pin was causing dependency conflicts for me.

Would it be possible to loosen the pin to allow users to manage their own datacube version if needed?

odc/stats/io.py
taufik-shf: You're right - it should be fixed in xr_reproject().

I'm happy to attempt a fix in odc-geo if you'd prefer, otherwise we could merge this workaround for now and I'll open an issue upstream.

odc/stats/plugins/gm.py
taufik-shf: yes, it should be fixed in geomedian_with_mads().

I can try to fix it in odc-algo if you'd prefer, or we can merge this workaround for now until the upstream is fixed. Once it is fixed there, I'll remove the workaround to keep the code clean.

odc/stats/model.py
pjonsson: Oh, I wasn't aware that is how it works in this repository. Well, it's probably best to keep the suppression then.

odc/stats/io.py
pjonsson: Without speaking for this repository, I think it would make sense to try to fix the problems in odc-algo and odc-geo. It sounds like you've found some bugs, and if I'm wrong and the methods are meant to behave the way they currently behave, you're more likely to get that feedback if you make PRs to those repositories.

odc/stats/plugins/gm.py
Kirill888: see https://github.com/opendatacube/odc-geo/issues/259#issuecomment-4044299679

PR #238, opened 2026-02-21
Convert setup.py to pyproject.toml

pjonsson:
The same files are included before/after
when doing "python3 -m build" on
my machine.


tests/requirements.txt
```diff
@@ -15,14 +15,14 @@ odc-algo>=1.0.1,<1.1 # No geomad yet
...
-odc-stac>=0.4.0
+odc-stac==0.4.0 # 0.4.1 removes stac2ds, not moved to datacube until 1.9.6
 
 # For tests
 pytest
 pytest-cov
 pytest-httpserver
 pytest-timeout
-s3fs
+s3fs>2023.6.0
```
pjonsson, line 25: Not sure what happened but the tests kept finding some ancient version of s3fs in the Conda cache.

The other changes to this file is alignment with the requirements in `docker/requirements.txt`.

Dockerfile
```diff
@@ -31,10 +31,7 @@ FROM ubuntu:jammy-20240212
 COPY --link --from=stats-conda /env /env
 COPY --link docker/distributed.yaml  /etc/dask/
 
-ENV GDAL_DRIVER_PATH=/env/lib/gdalplugins \ 
-    PROJ_LIB=/env/share/proj \
```
pjonsson, line 35: Noticed a warning in the tests about the PROJ database being of the wrong version, so removed these (each package needs to use the PROJ database they were compiled with).

.github/workflows/main.yml
```diff
@@ -89,15 +80,15 @@ jobs:
...
+          apt-get update
           DEBIAN_FRONTEND=noninteractive apt-get install -y \
           --fix-missing --no-install-recommends \
           curl \
           gpg \
           nodejs \
           zstd
           python -m pip install --upgrade --no-cache pip
-          pip install --upgrade --no-cache -r \
+          pip install --no-cache -r \
```
pjonsson, line 94: I'm fairly sure the `--upgrade` was masking the ancient s3fs version being installed, so I have removed those switches for everything except pip.

.github/workflows/create-new-tag.yml
pjonsson, line 1: This workflow looks like it's used for patching the package version. Switched to using setuptools-scm like the other Datacube packages.

andrewdhicks: 👍

Dockerfile
andrewdhicks: We were seeing errors in the docker image when pip pulled in `rasterio` via a dep in requirements.txt. The rasterio from pip is now bundled with its own copy of proj and gdal. This conflicted with the proj installed by micromamba.

I believe these lines are needed for the conda-forge versions, as in the final docker image we do not call `micromamba activate` on the environment for them to be set automatically. Happy to remove them for now, as I intend to address the conda/pip issue seperately.

andrewdhicks (approved):
Thanks Peter! I'll remember to check the PRs before I dive into a change again.

PR #243, opened 2026-05-20
Update stats 2026

andrewdhicks:
Aims of this release

- [ ] Align pyproject and versioning to other ODC apps
- [ ] Align github actions and releases to other ODC apps
- [ ] Update to newer python, datacube, and `odc` libraries
- [ ] Move away from any packages on packages.dea.ga.gov.au
- [ ] Fix test failures around bit-masking
- [ ] Resolve performance issues with masking


odc/stats/io.py
```diff
@@ -585,7 +585,7 @@ def choose_transform_path(
     dst_crs: str,
     transform_code: str | None = None,
     area_of_interest: Sequence[float] | None = None,
-) -> str:
+) -> dict[str, str]:
```
pjonsson, line 588: The commit fixes a type annotation, not a doc string.

PR #248, opened 2026-07-07
fix: utcnow deprecated in python 3.12

andrewdhicks:
To resolve:
```DeprecationWarning: datetime.datetime.utcnow() is deprecated and scheduled for removal in a future version. Use timezone-aware objects to represent datetimes in UTC: datetime.datetime.now(datetime.UTC).```

### DIFF
--- a/odc/stats/model.py
+++ b/odc/stats/model.py
@@ -2,7 +2,7 @@ import math
 from abc import ABC, abstractmethod
 from copy import deepcopy
 from dataclasses import dataclass, field
-from datetime import datetime, timedelta, timezone
+from datetime import datetime, timedelta, timezone, UTC
 from typing import Any
 from collections.abc import Sequence
 from uuid import UUID, uuid5
@@ -236,7 +236,7 @@ class WorkTokenInterface(ABC):
         """
         Implementations should use this method internally
         """
-        return datetime.utcnow()
+        return datetime.now(UTC)
 
     @property
     @abstractmethod
@@ -458,7 +458,7 @@ class Task:
         warnings.filterwarnings("default")
 
         if processing_dt is None:
-            processing_dt = datetime.utcnow()
+            processing_dt = datetime.now(tz=UTC)
         dataset_assembler.processed = processing_dt
 
         dataset_assembler.maturity = self.product.maturity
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

**Review by SpacemanPaul — COMMENTED** [review]
Some Github actions obviously need to be moved off Python 3.10.

Otherwise looks good.

File: tests/test_sqs.py [hunk 1]
```diff
@@ -42,8 +42,8 @@ def test_sqs_work_token(sqs_message):
     tk = SQSWorkToken(sqs_message, 60)
 
     assert tk.active_seconds < 2
-    assert tk.start_time < datetime.utcnow()
-    assert tk.deadline > datetime.utcnow()
+    assert tk.start_time < datetime.now(UTC)
+    assert tk.deadline > datetime.now(UTC)
```
SpacemanPaul: Could avoid calling `datetime.now()` twice by using the walrus operator `:=`

odc/stats/model.py
```diff
@@ -236,7 +236,7 @@ def now():
         """
         Implementations should use this method internally
         """
-        return datetime.utcnow()
+        return datetime.now(timezone.UTC)
```
pjonsson: I think this will pass the CI on Python 3.10 with lower-case UTC, so `timezone.utc`.

odc/stats/proc.py
```diff
@@ -195,7 +195,7 @@ def _register_heartbeat(self, hearbeat_filepath: str):
         Records the timestamp at which a hearbeat was detected
 
         """
-        t_now = datetime.utcnow()
+        t_now = datetime.now(timezone.UTC)
```
pjonsson: CI doesn't seem to catch this one for some reason.

tests/__init__.py
```diff
@@ -104,7 +104,7 @@ def mk_dask_xx(
     if chunks is None:
         chunks = {"x": -1, "y": -1}
     if timestamps is None:
-        timestamps = [datetime.utcnow()]
+        timestamps = [datetime.now(timezone.UTC)]
```
pjonsson: And here's another one that needs to be lower-cased as well.

pjonsson (approved):

PR #249, opened 2026-07-15
Replace gdal/ogr with geopandas

andrewdhicks:
Attempt at a version that only uses `rasterio` (with bundled `gdal` binary) from `pypi`, and not `gdal` from `conda-forge`.

### DIFF
--- a/odc/stats/_cli_locate_grids.py
+++ b/odc/stats/_cli_locate_grids.py
@@ -27,11 +27,11 @@ def locate_grids(grid_shape, extent_shape, attr_filter=None):
         grids = grids.to_crs(extents.crs)
 
     extent_geom = extents.geometry.union_all()
-    matched_grids = grids[grids.geometry.intersects(extent_geom)]
+    matched_grids = grids[grids.geometry.intersects(extent_geom)]["region_code"]
 
     return [
         re.findall(r"\d+", str(region_code))
-        for region_code in matched_grids["region_code"]
+        for region_code in matched_grids
     ]
 
 
--- a/odc/stats/_cli_locate_grids.py
+++ b/odc/stats/_cli_locate_grids.py
@@ -29,10 +29,7 @@ def locate_grids(grid_shape, extent_shape, attr_filter=None):
     extent_geom = extents.geometry.union_all()
     matched_grids = grids[grids.geometry.intersects(extent_geom)]["region_code"]
 
-    return [
-        re.findall(r"\d+", str(region_code))
-        for region_code in matched_grids
-    ]
+    return [re.findall(r"\d+", str(region_code)) for region_code in matched_grids]
 
 
 @main.command("locate-grids")
--- a/odc/stats/_cli_locate_grids.py
+++ b/odc/stats/_cli_locate_grids.py
@@ -18,13 +18,13 @@ def locate_grids(grid_shape, extent_shape, attr_filter=None):
     else:
         extents = gpd.read_file(extent_shape)
 
+    if extents.empty:
+        return []
+
     # Match CRS.
     if grids.crs is not None and extents.crs is not None and grids.crs != extents.crs:
         # We transform the grid as it can be in geographic projection with antemeridian issues
         grids = grids.to_crs(extents.crs)
-
-    if extents.empty:
-        return []
 
     extent_geom = extents.geometry.union_all()
     matched_grids = grids[grids.geometry.intersects(extent_geom)]
--- a/odc/stats/_cli_locate_grids.py
+++ b/odc/stats/_cli_locate_grids.py
@@ -4,34 +4,35 @@ import tempfile
 import pandas as pd
 from os import path
 from io import StringIO
-from osgeo import ogr, osr
+import geopandas as gpd
 from ._cli_common import main
 
 
-def locate_grids(grid_shape, extent_shape, attr_filter):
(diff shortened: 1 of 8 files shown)


odc/stats/_cli_locate_grids.py
```diff
@@ -4,34 +4,35 @@
...
+    # Match CRS.
+    if grids.crs is not None and extents.crs is not None and grids.crs != extents.crs:
+        # We transform the grid as it can be in geographic projection with antemeridian issues
+        grids = grids.to_crs(extents.crs)
+
+    if extents.empty:
+        return []
+
+    extent_geom = extents.geometry.union_all()
+    matched_grids = grids[grids.geometry.intersects(extent_geom)]
```
pjonsson: Put the `["region_code"]` in the RHS here instead so less of `grids` is kept alive and the lookup isn't done in the loop.

odc/stats/_cli_locate_grids.py
```diff
@@ -4,34 +4,35 @@
...
 from ._cli_common import main
 
 
-def locate_grids(grid_shape, extent_shape, attr_filter):
-    ds_grid = ogr.Open(grid_shape)
-    lyr_grid = ds_grid.GetLayer(0)
-    lyr_grid.ResetReading()
-    grid_spatial_ref = lyr_grid.GetSpatialRef()
+def locate_grids(grid_shape, extent_shape, attr_filter=None):
+    grids = gpd.read_file(grid_shape)
```
pjonsson, line 12: Does `read_file` open and close the file internally before returning, or do these need to be closed in some other way? A quick search doesn't give me much, but the example with fsspec at https://geopandas.org/en/stable/docs/user_guide/io.html#reading-spatial-data suggests putting it in an with-statement:
```python
path = "simplecache::http://download.geofabrik.de/antarctica-latest-free.shp.zip"
with fsspec.open(path) as file:
    df = geopandas.read_file(file)
```

odc/stats/_cli_locate_grids.py
```diff
@@ -4,34 +4,35 @@
...
-    for grid in lyr_grid:
-        grid_geom = grid.geometry()
-        grid_geom.Transform(transform)
-        lyr_extent.SetSpatialFilter(grid_geom)
-        if lyr_extent.GetFeatureCount() > 0:
-            extent_grids += [re.findall(r"\d+", grid["region_code"])]
-        lyr_extent.SetSpatialFilter(None)
-        lyr_extent.ResetReading()
-    return extent_grids
+        extents = gpd.read_file(extent_shape, where=attr_filter)
```
pjonsson, line 17: I just noticed fsspec is a project dependency already, should these be opened through fsspec to support more than just local file storage, or does that not matter for this part of the code?

odc/stats/_cli_locate_grids.py
```diff
@@ -4,34 +4,35 @@
...
+    else:
+        extents = gpd.read_file(extent_shape)
+
+    # Match CRS.
+    if grids.crs is not None and extents.crs is not None and grids.crs != extents.crs:
+        # We transform the grid as it can be in geographic projection with antemeridian issues
+        grids = grids.to_crs(extents.crs)
+
+    if extents.empty:
```
pjonsson: Shouldn't this check come right after `extents = ...`?

odc/stats/plugins/_utils.py
```diff
@@ -1,51 +1,35 @@
...
-
     if threshold is not None:
         if mask.sum() > mask.size * threshold:
-            return dask.array.ones(dst_shape, name=False)
+            return da.ones(geobox.shape.shape, name=False)
 
-    return dask.array.from_array(mask.reshape(dst_shape), name=False)
+    return da.from_array(
+        mask.reshape(geobox.shape.shape),
+        name=False,
```
pjonsson: If you remove the final comma I think Ruff will reformat this into a single line just like the old code.

odc/stats/plugins/_utils.py
```diff
@@ -1,51 +1,35 @@
...
-        transform.c,
-        transform.a,
-        transform.b,
-        transform.f,
-        transform.d,
-        transform.e,
+    gdf = gpd.read_file(shape_file, where=filter_expression)
+    geom = Geometry(
+        gdf.union_all(),
+        crs=gdf.crs,
```
pjonsson: If you remove the final comma I think Ruff will reformat this into a single line.

odc/stats/_cli_locate_grids.py
andrewdhicks: Internally it is doing the open/close using the appropriate drivers. I think the with statement is only appropriate for file-like objects.

odc/stats/_cli_locate_grids.py
andrewdhicks: OGR and geopandas support http, s3, zipfiles, etc the same way that GDAL does, so it is equivalent to the previous version.

It looks like fsspec is being used to read files that are written by hand-crafted file/S3 code in io.py.
That is a mess, but fixing that is outside the scope of this change.

odc/stats/_cli_locate_grids.py
andrewdhicks: Good idea!

SpacemanPaul (approved):
LGTM.  Some nice cleanup here.

